PR #9, opened 2021-05-27
Week1

dinithiW:
The week1 folder has the https://github.com/broadinstitute/warp/blob/develop/tasks/broad/Qc.wdl file and the manual translation of it using cwl. Also includes the errors I got while running the two files. The md5 folder has the md5 checksum workflow containing both wdl and cwl files and their respective outputs.


dinithiW:
Commands: 

WDL:
miniwdl check Qc.wdl --no-shellcheck  : to check for errors in .wdl file

miniwdl run Qc.wdl  input_bam=simple.bam  metrics_filename="metrics_file"  preemptible_tries=2 --verbose : run the workflow

CWL: 
cwltool --validate Qc.cwl
cwl-runner Qc.cwl Qc.yml

Errors:
The errors are in the cwl_error.txt and wdl_error.txt files inside the week1 folder. If you go in to the 20210527_195245_CollectQualityYieldMetrics/ folder, error.json gives more details on the error (for the wdl run)

Task2&3/WdlV1_1Ast.py
```diff
@@ -0,0 +1,4 @@
+class TaskNode():
```
mr-c, line 1: Lets move the WDL parser/lexer into its own folder. We don't need the Java code versions, so those can be deleted

Task2&3/sample.txt
```diff
@@ -0,0 +1,208 @@
+version 1.0
```
mr-c, line 1: Why is this a `.txt` file?

dinithiW: Actually it's not relevant to the work I did this week. These are from a long time back. Should have deleted them :/

dinithiW:
> Were you able to run `CollectQualityYieldMetrics` manually, without using WDL or CWL?

Did you mean using the tool? http://broadinstitute.github.io/picard/ this one?

Week1/Qc.cwl
```diff
@@ -0,0 +1,31 @@
...
+
+outputs:
+  quality_yield_metrics:
+    type: File
+    outputBinding:
+      glob: $(inputs.metrics_filename)
+
+baseCommand: [java, -Xms2000m, -jar, /usr/picard/picard.jar, CollectQualityYieldMetrics ]
+
+arguments: [INPUT=$(inputs.input_bam.path) ,OQ=true, OUTPUT=$(inputs.metrics_filename)]
```
mr-c, line 31:
```suggestion
arguments: [INPUT=$(inputs.input_bam.path), OQ=true, OUTPUT=$(inputs.metrics_filename)]
```

I wonder where the extra space is coming from?

Week1/create_cwl.py
```diff
@@ -0,0 +1,58 @@
...
+        cwl.ResourceRequirement(
+            ramMin=3584,
+        )
+    ]
+
+    outputs = [
+        cwl.CommandOutputParameter(
+            id="quality_yield_metrics",
+            type="File",
+            outputBinding=cwl.CommandOutputBinding(glob="$inputs.metrics_filename"),
```
mr-c, line 30:
```suggestion
            outputBinding=cwl.CommandOutputBinding(glob="$(inputs.metrics_filename)"),
```

dinithiW:
I have added the setup.py file and the modified python code(create_cwl.py) to create the cwl file (result.cwl)

dinithiW:
WdlV1_1ParserVisitor.py has been modified to get the inputs, outputs, and the relevant fields of a task required for the creation of the equivalent .cwl file.

dinithiW:
Converted “CollectQualityYieldMetrics” task to CWL using cwl-utils after parsing the wdl file

Week1/wdl_parsing/test_WdlV1_1.py
```diff
@@ -0,0 +1,159 @@
...
+#only to handle values given in GiB
+def get_ram_min(ram_min):
+    ram_min = ram_min[ram_min.find("\"")+1:ram_min.find("GiB")]
+    return int(float(ram_min.strip())*1024)
+
+import cwl_utils.parser_v1_2 as cwl
+
+from ruamel import yaml
+
+def main() -> None:
```
mr-c, line 95: I suggest adding a command line option to store the results.

Week1/wdl_parsing/test_WdlV1_1.py
```diff
@@ -0,0 +1,159 @@
...
+command_arguments = []
+
+for a in command:
+    #if it contains = , it's taken as an argument
+    if "=" in a:
+        command_arguments.append(a.strip())
+    #else add to base command
+    else:
+        base_command+=a
+
```
mr-c, line 63: This works, good!

1. `=` is not required for a CWL parameter reference or CWL expression. Search instead for `$(` or `${`
2. what happens if a later command argument does not contain a `=` ? Then it would be added to the `baseCommand` and that will affect the command line order: https://www.commonwl.org/v1.2/CommandLineTool.html#Input_binding

Week1/wdl_parsing/test_WdlV1_1.py
```diff
@@ -0,0 +1,159 @@
...
+    
+    inputs = []
+    for i in ast.task_inputs:
+        input_type = wdl_type[i[0]]
+        input_name = i[1]
+        inputs.append(cwl.CommandInputParameter(id=input_name, type=input_type))  
+
+    docker_requirement = [
+        cwl.DockerRequirement(
+            dockerPull=ast.task_runtime["docker"],
```
mr-c, line 106: Here you should remove any extra pairs of quotation marks

Week1/wdl_parsing/test_WdlV1_1.py
```diff
@@ -0,0 +1,159 @@
...
+        ))
+
+    arguments = []
+
+    for i in command_arguments:
+        arguments.append(cwl.CommandLineBinding(valueFrom=i))
+    
+
+    cat_tool = cwl.CommandLineTool(
+        id="CollectQualityYieldMetrics",
```
mr-c, line 141: Can you derive this `id` from the `WDL` file?

mr-c (changes requested):
Please move everything out of the `Week1` folder

Week1/wdl_parsing/test_WdlV1_1.py
dinithiW: but this won't work for an argument like 'OQ=true' as mentioned in this wdl file right?

dinithiW:
I have fixed the errors as discussed and moved everything out of the week1 folder.

mr-c (approved):

PR #11, opened 2021-06-10
add pytest

dinithiW:


mr-c (changes requested):
Can you look over https://docs.pytest.org/en/6.2.x/goodpractices.html#choosing-a-test-layout-import-rules so that you don't need to move the python files?

mr-c:
Lets name the package `wdl2cwl`

mr-c:
When you're done with the PR, I'll setup https://en.wikipedia.org/wiki/Continuous_integration with GitHub Actions to automatically run pytest on every PR after this one

tests/WdlV1_1.py
```diff
@@ -0,0 +1,146 @@
...
+index = 0
+for i in command_arguments:
+    if "~" in i:
+        parameter_reference = i[i.find("~{")+2:i.find("}")] #finding the name of the parameter
+        sub_str = i.strip().split("~")
+        if "INPUT" not in sub_str[0]:
+            command_arguments[index] = sub_str[0]+"$(inputs."+parameter_reference+")"
+        else:
+            command_arguments[index] = sub_str[0]+"$(inputs."+parameter_reference+".path"")"
+    index+=1
```
mr-c, line 51: Move this into the `main` function. Add an argument for the file name of the input WDL file

dinithiW:
> Lets name the package `wdl2cwl` and then follow https://docs.pytest.org/en/6.2.x/goodpractices.html#tests-as-part-of-application-code

@mr-c Do we have to change the project name to wdl2cwl as well? Anyway, having the name as wdl-cwl-translator gives a syntax error when used for imports

mr-c:
> > Lets name the package `wdl2cwl` and then follow https://docs.pytest.org/en/6.2.x/goodpractices.html#tests-as-part-of-application-code
> 
> @mr-c Do we have to change the project name to wdl2cwl as well? Anyway, having the name as wdl-cwl-translator gives a syntax error when used for imports

That makes sense, the `-` will be interpreted as a minus sign. So either use `wdl_cwl_translator` or `wdl2cwl`

mr-c (approved):

PR #14, opened 2021-06-16
allow running in place

mr-c:


dinithiW (commented):
New error when I run from the root.
Traceback (most recent call last):
  File "wdl2cwl/WdlV1_1.py", line 3, in <module>
    from wdl2cwl.WdlV1_1Lexer import WdlV1_1Lexer
ModuleNotFoundError: No module named 'wdl2cwl.WdlV1_1Lexer'

mr-c:
Try `python setup.py develop` first

dinithiW:
Yes it works!!

PR #18, opened 2021-06-29
Wdl meta

dinithiW:
Issue warning for parameter meta and meta.
Add test case for parameter_meta.


wdl2cwl/tests/cwl_files/smoove.cwl
```diff
@@ -0,0 +1,24 @@
...
+  type: File
+- id: referenceFastaFai
+  type: File
+- id: sample
+  type: string
+requirements:
+- class: InitialWorkDirRequirement
+  listing:
+  - entryname: example.sh
+    entry: 'set -e mkdir -p $(inputs.outputDir) smoove call  --outdir $(inputs.outputDir)  --name
```
mr-c, line 18: This should have been broken up into three separate lines, not smooshed all together 😃

mr-c (changes requested):

wdl2cwl/tests/cwl_files/smoove.cwl
mr-c: Let's do a video chat and we can debug this together

dinithiW:
The new command does not work with the input file used previously. Giving out an error
```
ERROR [job CollectQualityYieldMetrics] Job error:
("Error collecting output for parameter 'quality_yield_metrics':\nNewResult.cwl:12:5: Did not find output file with glob pattern: '['metrics_file']'", {})
WARNING [job CollectQualityYieldMetrics] completed permanentFail
{}
WARNING Final process status is permanentFail

File: wdl2cwl/main.py [hunk 2]
```diff
@@ -78,13 +80,14 @@ def get_command(
...
+
             index = end_index + 1
         else:
             if command[index] != "\\":
                 new_command += command[index]
                 index += 1
             else:
-                new_command += " "
+                # new_command += " "
                 index += 1
```
mr-c, line 91: I think this block is the problem, you're filtering out the trailing `\` which we want to keep
``` diff
@@ -83,12 +84,8 @@ def get_command(
 
             index = end_index + 1
         else:
-            if command[index] != "\\":
-                new_command += command[index]
-                index += 1
-            else:
-                # new_command += " "
-                index += 1
+            new_command += command[index]
+            index += 1
     return new_command
```

wdl2cwl/WdlV1_1ParserVisitor.py
```diff
@@ -56,28 +58,30 @@ def visitWdl_type(self, ctx:WdlV1_1Parser.Wdl_typeContext):
     # Visit a parse tree produced by WdlV1_1Parser#unbound_decls.
     def visitUnbound_decls(self, ctx:WdlV1_1Parser.Unbound_declsContext):
         input_type = self.visitWdl_type(ctx.wdl_type())
-        return [input_type, str(ctx.Identifier())]
+        if self.task_input_check:
+            self.task_inputs.append([input_type, str(ctx.Identifier())])
+        '''elif self.workflow_input_check:
+            self.workflow_inputs_unbound.append([input_type, str(ctx.Identifier())])'''
```
mr-c, line 64:
```suggestion
```
Lets not commit commented out code

mr-c (changes requested):

wdl2cwl/WdlV1_1ParserVisitor.py
```diff
@@ -356,18 +354,17 @@ def visitMeta_kv(self, ctx:WdlV1_1Parser.Meta_kvContext):
...
     # Visit a parse tree produced by WdlV1_1Parser#parameter_meta.
     def visitParameter_meta(self, ctx:WdlV1_1Parser.Parameter_metaContext):
-        return self.visitChildren(ctx)
+        self.task_parameter_meta_check = 1
 
 
     # Visit a parse tree produced by WdlV1_1Parser#meta.
     def visitMeta(self, ctx:WdlV1_1Parser.MetaContext):
-        return self.visitChildren(ctx)
+        self.task_meta_check = 1
```
mr-c, line 362: This line is never called by any of the tests. Can you either add a meta section to one of the tests, or remove this?

PR #22, opened 2021-07-06
Handle bound inputs

dinithiW:
Handle bound inputs
Convert int, boolean type values of bound inputs from string to their respective type.

### DIFF
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -90,6 +90,35 @@ def get_command(
             index += 1
     return new_command
 
+def get_input(inputs,unbound_input,bound_input):
+    """Get bound and unbound inputs."""
+    for i in unbound_input:
+        input_type = wdl_type[i[0]]
+        input_name = i[1]
+        inputs.append(
+            cwl.CommandInputParameter(
+                id=input_name, 
+                type=input_type
+                )
+            )
+
+    for i in bound_input:
+        input_type = wdl_type[i[0]]
+        input_name = i[1]
+        input_value = i[2].replace('"',"")
+
+        if input_type=='boolean':
+            input_value = bool(input_value)
+
+        inputs.append(
+            cwl.CommandInputParameter(
+                id=input_name,
+                type=input_type,
+                default=input_value,
+            )
+        )
+    
+    return inputs
 
 def main(argv: List[str]) -> str:
     """Generate a CWL object to match "cat-tool.cwl"."""
@@ -129,15 +158,11 @@ def main(argv: List[str]) -> str:
     command = get_command(
         command, ast.task_inputs, ast.task_inputs_bound, input_types, input_names
     )
-    print(command)
 
     base_command = ["sh", "example.sh"]
 
     inputs = []
-    for i in ast.task_inputs:
-        input_type = wdl_type[i[0]]
-        input_name = i[1]
-        inputs.append(cwl.CommandInputParameter(id=input_name, type=input_type))
+    inputs = get_input(inputs,ast.task_inputs,ast.task_inputs_bound)
 
     requirements: List[cwl.ProcessRequirement] = []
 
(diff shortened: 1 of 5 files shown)


wdl2cwl/main.py
```diff
@@ -91,6 +91,38 @@ def get_command(
...
+    """Get bound and unbound inputs."""
+    for i in unbound_input:
+        input_type = wdl_type[i[0]]
+        input_name = i[1]
+        inputs.append(cwl.CommandInputParameter(id=input_name, type=input_type))
+
+    for i in bound_input:
+        input_type = wdl_type[i[0]]
+        input_name = i[1]
+        input_value: Any = i[2].replace('"', "")
```
mr-c, line 108: Why `Any` here?

wdl2cwl/tests/cwl_files/seqtk.cwl
```diff
@@ -0,0 +1,42 @@
...
+
+            set -e -o pipefail
+            mkdir -p "$(dirname $(inputs.outFilePath))"
+            $(inputs.preCommand)
+            seqtk sample \
+            $(inputs."-s "+seed) \
+            $(inputs.true="-2 "false=""twoPassMode) \
+            $(inputs.sequenceFile.path) \
+            $(inputs.fractionOrNumber) \
+            $(inputs.true="| gzip"false=""zip) \
```
mr-c, line 36: These aren't valid CWL expressions. As usual, you can either
1) simplfy your test to not include these advanced cases & make an issue to come back to this later
2) Fix the advanced cases now in this PR

mr-c (changes requested):

wdl2cwl/main.py
dinithiW: When I run tox I got an incompatible type assignment error, because the value of input_value would later be changed to boolean and int.

mr-c: That's fine, you can notate the multiple types it can be like this

```suggestion
        input_value: Union[str, bool, int] = i[2].replace('"', "")
```

dinithiW:
```
wdl2cwl/main.py:113:32: error: Item "bool" of "Union[str, bool, int]" has no attribute "lower"  [union-attr]
                input_value = bool(input_value.lower() == "true")
                                   ^
wdl2cwl/main.py:113:32: error: Item "int" of "Union[str, bool, int]" has no attribute "lower"  [union-attr]
                input_value = bool(input_value.lower() == "true")
```
Now it gives this error. Shall I switch back to Any?

mr-c: There are several possible solutions. The simplest may be to have a `raw_input_value` of the type `str` on line 110 and an `input_value` of type `Union[str, bool, int]` used for the lines 112-115 section, turned that section into an `if` boolean, `elif` int, `elif` string, `else` throw an exception block.

wdl2cwl/tests/cwl_files/seqtk.cwl
```diff
@@ -0,0 +1,39 @@
...
+        glob: outFilePath
+requirements:
+  - class: InitialWorkDirRequirement
+    listing:
+      - entryname: example.sh
+        entry: |4
+
+            set -e -o pipefail
+            mkdir -p "$(dirname $(inputs.outFilePath))"
+            $(inputs.preCommand)
```
mr-c, line 30: Have you tried running this CWL description yet?

dinithiW: No. I have only run CollectQualityYieldMetrics.cwl yet.

mr-c: okay. Please test it. I think the `$(dirname ` might be mistaken for a CWL expression or parameter reference and may need to be escaped

PR #24, opened 2021-07-08
Handle WDL requirements

dinithiW:
Added WDL requirement time minutes.
Modified docker and memory requirement to get the value from bound inputs in case the requirements are not defined in the requirements section as a string.

### DIFF
--- a/wdl2cwl/tests/cwl_files/smoove.cwl
+++ b/wdl2cwl/tests/cwl_files/smoove.cwl
@@ -12,6 +12,8 @@ inputs:
   - id: sample
     type: string
 requirements:
+  - class: DockerRequirement
+    dockerPull: quay.io/biocontainers/smoove:0.2.5--0
   - class: InitialWorkDirRequirement
     listing:
       - entryname: example.sh
@@ -25,6 +27,11 @@ requirements:
             --fasta $(inputs.referenceFasta.path) \
             $(inputs.bamFile.path)
   - class: InlineJavascriptRequirement
+hints:
+  - class: ResourceRequirement
+    ramMin: 15360
+  - class: ToolTimeLimit
+    timelimit: 86400
 cwlVersion: v1.2
 baseCommand:
   - sh
--- a/wdl2cwl/tests/cwl_files/smoove.cwl
+++ b/wdl2cwl/tests/cwl_files/smoove.cwl
@@ -27,11 +27,10 @@ requirements:
             --fasta $(inputs.referenceFasta.path) \
             $(inputs.bamFile.path)
   - class: InlineJavascriptRequirement
-hints:
   - class: ResourceRequirement
     ramMin: 15360
   - class: ToolTimeLimit
-    timelimit: 86400
+    timelimit: $(inputs.timeMinutes* 60)
 cwlVersion: v1.2
 baseCommand:
   - sh
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -197,8 +197,6 @@ def main(argv: List[str]) -> str:
             for sublist in ast.task_inputs_bound:
                 if sublist[1] in ast.task_runtime["time_minutes"]:
                     time_minutes = sublist[2]
-        else:
-            time_minutes = ast.task_runtime["time_minutes"]
 
         hints.append(
             cwl.ToolTimeLimit(
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -38,11 +38,6 @@ def get_ram_min(ram_min: str) -> int:
     return int(float(ram_min.strip()) * 1024)
 
 
-def get_time_minutes(time_minutes: str) -> int:
-    """Convert minutes to seconds."""
-    return int(float(time_minutes.strip()) * 60)
-
-
 def get_command(
     command: str,
     unbound: List[str],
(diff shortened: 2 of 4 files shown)


wdl2cwl/tests/cwl_files/smoove.cwl
```diff
@@ -25,6 +27,11 @@ requirements:
             --fasta $(inputs.referenceFasta.path) \
             $(inputs.bamFile.path)
   - class: InlineJavascriptRequirement
+hints:
+  - class: ResourceRequirement
+    ramMin: 15360
+  - class: ToolTimeLimit
+    timelimit: 86400
```
mr-c, line 34: In the original WDL, the time limit was a user parameter with a default value, so we would do the same here

wdl2cwl/tests/cwl_files/smoove.cwl
```diff
@@ -25,6 +27,11 @@ requirements:
             --fasta $(inputs.referenceFasta.path) \
             $(inputs.bamFile.path)
   - class: InlineJavascriptRequirement
+hints:
```
mr-c, line 30: Why `hints` instead of `requirements`? The WDL `runtime` directive is not optional: https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#runtime-section

mr-c (changes requested):

wdl2cwl/tests/cwl_files/smoove.cwl
dinithiW: I got that default value and converted into seconds. Is that incorrect?

wdl2cwl/tests/cwl_files/smoove.cwl
dinithiW: Ok. I will move it in to requirements

wdl2cwl/tests/cwl_files/smoove.cwl
mr-c: `timelimit` should be an expression referring to a input with that default value, just like in the original https://github.com/common-workflow-lab/wdl-cwl-translator/pull/24/files#diff-3403794b7459aa2221c7a16897161d454dddce70494206e5009454ce46750e3dR33

dinithiW: I'm not sure whether I understood correcly. ```timelimit``` should be something like $(inputs.timeMinutes) and doesn't need to be converted to seconds?

mr-c: Probably safest to keep the original default value and do the conversion using a CWL expression `$(inputs.timeMinutes  * 60)`

dinithiW: Ok. Can I do the same for other requirements (eg: docker requirement) too? Use the expression instead of replacing with the default value? ```$(inputs.docker)``` instead of replacing it with the actual value.

mr-c: To answer that question, we check the standard.

https://www.commonwl.org/v1.2/CommandLineTool.html#DockerRequirement

Alas, only a static string is accepted for any of the fields in `dockerRequirement`

dinithiW: Shall I convert the memory (ResourceRequirement: ramMin) requirement to an expression as well or keep it as it is?

wdl2cwl/tests/cwl_files/smoove.cwl
```diff
@@ -25,6 +27,10 @@ requirements:
             --fasta $(inputs.referenceFasta.path) \
             $(inputs.bamFile.path)
   - class: InlineJavascriptRequirement
+  - class: ResourceRequirement
+    ramMin: 15360
+  - class: ToolTimeLimit
+    timelimit: $(inputs.timeMinutes* 60)
```
mr-c, line 33: `timeMinutes` is missing from the inputs?

mr-c (changes requested):

wdl2cwl/tests/cwl_files/smoove.cwl
dinithiW: This was the one that was discussed yesterday for which you suggested to come up with a javascript function.

dinithiW: I don't follow. It is available in the inputs section with a default value.

mr-c: `timeMinutes` only occurs once in the generated CWL file that you checked in: https://github.com/common-workflow-lab/wdl-cwl-translator/blob/bed19b13188fb20b918b90b52eaf99598a20f101/wdl2cwl/tests/cwl_files/smoove.cwl

wdl2cwl/tests/cwl_files/smoove.cwl
mr-c: Yes, please also convert `ramMin` into an expression

wdl2cwl/tests/cwl_files/smoove.cwl
dinithiW: oh yes. It's available in the inputs section as a bound input, which are addressed in the other PR #22 and not in the PR.

mr-c: That makes more sense 👍🏻. I've merged that PR, so please update this branch with the latest from `main`.

mr-c (approved):

PR #29, opened 2021-07-14
Optional inputs

dinithiW:
Handle optional (?) inputs

### DIFF
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -90,6 +90,35 @@ def get_command(
             index += 1
     return new_command
 
+def get_input(inputs,unbound_input,bound_input):
+    """Get bound and unbound inputs."""
+    for i in unbound_input:
+        input_type = wdl_type[i[0]]
+        input_name = i[1]
+        inputs.append(
+            cwl.CommandInputParameter(
+                id=input_name, 
+                type=input_type
+                )
+            )
+
+    for i in bound_input:
+        input_type = wdl_type[i[0]]
+        input_name = i[1]
+        input_value = i[2].replace('"',"")
+
+        if input_type=='boolean':
+            input_value = bool(input_value)
+
+        inputs.append(
+            cwl.CommandInputParameter(
+                id=input_name,
+                type=input_type,
+                default=input_value,
+            )
+        )
+    
+    return inputs
 
 def main(argv: List[str]) -> str:
     """Generate a CWL object to match "cat-tool.cwl"."""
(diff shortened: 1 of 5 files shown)


dinithiW:
Handle array type inputs

dinithiW:
https://github.com/biowdl/tasks/blob/develop/rtg.wdl#L23

mr-c (changes requested):
Lets keep the array code for a different pull request. Remove that and I'll merge this

wdl2cwl/tests/cwl_files/seqtk.cwl
```diff
@@ -5,6 +5,14 @@ inputs:
     type: File
   - id: fractionOrNumber
     type: float
+  - id: preCommand
+    type:
+      - string
+      - 'null'
```
mr-c, line 11: If we don't provide a "preCommand" what goes in the bash script for `$(inputs.preCommand)` ? Is it an empty string or the word "null"? 

If it is the later, then we will have a problem and we should set the default value for optional strings to `""`

mr-c: @dinithiW can you respond to my question above?

dinithiW: I didn't quite get back this PR again. I tried running with a Fasta file to which I got a command failed error

mr-c: Can you share the error? More importantly, can you run with `cwltool --cachedir some_path wdl2cwl/tests/cwl_files/seqtk.cwl ` and share the rendered `example.sh`?

dinithiW: resultant example.sh
```
set -e -o pipefail
        mkdir -p "$(dirname subsampledReads.fq.gz)"
        null
        seqtk sample 
	    -s null 
        /tmp/77vey6du/stg5cc2fcb9-f851-4728-ae19-afa32d16efaf/insert_size_metrics_test.sam 
        1.0 
        >  subsampledReads.fq.gz
```

mr-c: Okay, just as I feared.

To quote myself:

> If it is the later, then we will have a problem and we should set the default value for optional strings to `""`

dinithiW: Ok. I will find a workaround for that.

wdl2cwl/main.py
```diff
@@ -87,16 +87,17 @@ def get_command(
...
+                append_str = (
+                    '$(inputs["'
+                    + sub_str
+                    + '"]===null?"":inputs["'
+                    + sub_str
+                    + '"]["class"]==="File"? "$(inputs.'
+                    + sub_str
+                    + '.path)": "$(inputs.'
+                    + sub_str
+                    + ')")'
```
mr-c, line 99: Instead of using Javascript here, set the default value to `""` when you construct a cwl.CommandInputParameter for an optional input string that has now default value of its own

dinithiW: Shall I keep '.path' for files or remove it?

dinithiW:
Removed JS and set the default value of optional inputs without a default value to ""

mr-c (approved):

Issue #27, opened 2021-07-08
Concatenations in command are not handled


dinithiW:
eg: ~{"-s " + seed}

A use of the concatenation can be found in https://github.com/biowdl/tasks/blob/develop/seqtk.wdl#L40

mr-c:
Basic implementation done in #30

PR #30, opened 2021-07-15
Handle string concatenations in command

dinithiW:
#27 is addressed here

### DIFF
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -3,7 +3,7 @@ import argparse
 from argparse import Namespace
 import sys
 from io import StringIO
-from typing import List, cast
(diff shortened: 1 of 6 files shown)


wdl2cwl/main.py
```diff
@@ -72,19 +72,35 @@ def get_command(
...
+
+                for i in split_str:
+                    if '"' in i:
+                        new_command += i.replace('"', "")
+                    else:
+                        index = input_names.index(i)
+                        data_type = input_types[index]  # get the data type of the input
+
+                        if data_type == "File":
+                            new_command += "$(inputs." + i + ".path)"
```
mr-c, line 87: Please find a test that uses this line

Issue #25, opened 2021-07-08
Expression Placeholder Option 'true and false' is not handled


dinithiW:
e.g. ~{true="--yes" false="--no" boolean_value}
https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#-expression-placeholder-options mentions that expression placeholder options are deprecated and will be removed in WDL 2.0.

https://github.com/biowdl/tasks/blob/develop/seqtk.wdl#L41 uses the mentioned expression placeholder option.

mr-c:
Fixed in https://github.com/common-workflow-lab/wdl-cwl-translator/pull/37

PR #37, opened 2021-07-22
Handle expression placeholder true and false

dinithiW:
Handles #25


wdl2cwl/tests/cwl_files/seqtk.cwl
```diff
@@ -34,8 +34,24 @@ requirements:
...
                     seqtk sample \
             	    -s $(inputs.seed) \
+                    ${var value="";
+            if(inputs["twoPassMode"]){
+            value="-2 ";
+            }
+            else{
+            value="";
+            }
+            return value;} \
```
mr-c, line 44:
```suggestion
                    ${return inputs["twoPassMode"] ? "-2" : "";
```
Can you adjust the template to use this more concise javascript syntax?

PR #40, opened 2021-07-27
Handle array inputs

### LINKED ISSUE #34: Array type inputs are not handled
Currently, array type WDL inputs are not translated. 
An array type input can be found here. https://github.com/biowdl/tasks/blob/develop/rtg.wdl#L25

dinithiW:
~~#34 is addressed here~~

Fixes #34 
(Writing it this was means that merging this PR will close that issue)

### DIFF
--- a/wdl2cwl/tests/wdl_files/hisat2.wdl
+++ b/wdl2cwl/tests/wdl_files/hisat2.wdl
@@ -1,4 +1,5 @@
 version 1.0
+# Source: https://github.com/biowdl/tasks/blob/develop/hisat2.wdl
 
 # Copyright (c) 2017 Leiden University Medical Center
 #
--- a/wdl2cwl/tests/wdl_files/hisat2.wdl
+++ b/wdl2cwl/tests/wdl_files/hisat2.wdl
@@ -57,7 +57,7 @@ task Hisat2 {
         mkdir -p "$(dirname ~{outputBam})"
         hisat2 \
         -p ~{threads} \
-        -x ~{sub(indexFiles[0], "\.[0-9]\.ht2", "")} \
+        
         ~{true="-1" false="-U" defined(inputR2)} ~{inputR1} \
         ~{"-2" + inputR2} \
         --rg-id ~{readgroup} \
@@ -77,7 +77,7 @@ task Hisat2 {
 
     output {
         File bamFile = outputBam
-        File summaryFile = summaryFilePath
+        #File summaryFile = summaryFilePath
     }
 
     runtime {
--- a/wdl2cwl/tests/wdl_files/hisat2.wdl
+++ b/wdl2cwl/tests/wdl_files/hisat2.wdl
@@ -1,0 +1,114 @@
+version 1.0
+
+# Copyright (c) 2017 Leiden University Medical Center
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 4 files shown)


wdl2cwl/tests/wdl_files/hisat2.wdl
```diff
@@ -0,0 +1,114 @@
...
+    Int estimatedSortThreads = if threads == 1 then 1 else 1 + ceil(threads / 4.0)
+    Int totalSortThreads = select_first([sortThreads, estimatedSortThreads])
+    Int estimatedMemoryGb = 1 + ceil(size(indexFiles, "G") * 1.2) + sortMemoryPerThreadGb * totalSortThreads
+
+    command {
+        set -e -o pipefail
+        mkdir -p "$(dirname ~{outputBam})"
+        hisat2 \
+        -p ~{threads} \
+
```
mr-c, line 60: Was there originally a blank line here? This doesn't seem right..

wdl2cwl/tests/wdl_files/hisat2.wdl
```diff
@@ -0,0 +1,114 @@
+version 1.0
+
+# Copyright (c) 2017 Leiden University Medical Center
```
mr-c, line 3: What URL is this file from?

mr-c (changes requested):
In the example, I'm surprised to not see `indexFiles` mentioned in the `command` block; is something missing?

wdl2cwl/tests/wdl_files/hisat2.wdl
dinithiW: No. I removed that line (```-x ~{sub(indexFiles[0], "\.[0-9]\.ht2", "")} \```) from the command because it contains the sub() function which is not yet implemented. I have to write a js expression for that which I thought could go to a separate PR. Shall I implement it in this PR itself?

wdl2cwl/tests/wdl_files/hisat2.wdl
dinithiW: I will add that

wdl2cwl/tests/wdl_files/hisat2.wdl
mr-c: Got it. For now you could use a simpler expression that does not use `sub` like `-x ~{indexFiles[0]}` with a note that it won't be valid command line for `hisat2`. Later when you implement `sub` you can use the original expression and remove the note.

wdl2cwl/tests/wdl_files/hisat2.wdl
```diff
@@ -1,4 +1,5 @@
 version 1.0
+# Source: https://github.com/biowdl/tasks/blob/develop/hisat2.wdl
```
mr-c, line 2: It is a bit safer to use the "permalink" version. On almost any GitHub page, if you press the `y` key on your keyboard the URL will change to a permanent version. In this case it is https://github.com/biowdl/tasks/blob/bc1bacf11498d2d30b85591cfccdcf71ef0966a5/hisat2.wdl

mr-c (changes requested):

dinithiW:
Changed source link of the WDL file.
Simplified commands in the WDL file and updated the corresponding CWL file.

mr-c (approved):

PR #47, opened 2021-08-03
Add defined() function

### LINKED ISSUE #44: WDL function defined() is not handled
Currently, the WDL function defined() is not translated by the convertor. (https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#boolean-definedx)

A WDL workflow containing defined() function can be found here. (https://github.com/biowdl/tasks/blob/bc1bacf11498d2d30b85591cfccdcf71ef0966a5/hisat2.wdl#L61)

dinithiW:
Fixes #44

### DIFF
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -136,6 +136,19 @@ def get_command(
                     )
                     new_command += append_str
 
+                elif "defined(" in input_name:
+                    sub_str = input_name[input_name.find("(") + 1 : -1]
+                    append_str = (
+                        '$(inputs["'
+                        + sub_str
+                        + '"] ? '
+                        + true_value
+                        + ' : "'
+                        + false_value
+                        + '")'
+                    )
+                    new_command += append_str
+
             else:
 
                 data_type = (
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -42,12 +42,18 @@ def get_ram_min(ram_min: str) -> int:
     return ram_value
 
 
-def get_ram_min_js(ram_min: str) -> str:
+def get_ram_min_js(ram_min: str, unit: str) -> str:
     """Get memory requirement for user input."""
+    append_str: str = ""
+    if unit:
+        append_str = '${\nvar unit = "' + unit + '";'
+    else:
+        append_str = (
+            '${\nvar unit = inputs["' + ram_min + '"].match(/[a-zA-Z]+/g).join("");'
+        )
     js_str = (
-        '${\nvar unit = inputs["'
-        + ram_min
-        + '"].match(/[a-zA-Z]+/g).join("");\nvar value = parseInt(inputs["'
+        append_str
+        + '\nvar value = parseInt(inputs["'
         + ram_min
         + '"].match(/[0-9]+/g));\n'
         + 'var memory = "";\n'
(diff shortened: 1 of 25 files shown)


wdl2cwl/main.py
```diff
@@ -136,6 +136,19 @@ def get_command(
...
+                elif "defined(" in input_name:
+                    sub_str = input_name[input_name.find("(") + 1 : -1]
+                    append_str = (
+                        '$(inputs["'
+                        + sub_str
+                        + '"] ? '
+                        + true_value
+                        + ' : "'
+                        + false_value
+                        + '")'
```
mr-c, line 148:
```suggestion
                    append_str = (
                        '$(inputs["'
                        + sub_str
                        + '" !== null] ? '
                        + true_value
                        + ' : "'
                        + false_value
                        + '")'
```

In OpenWDL 1.1, `defined()` is specified at https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#boolean-definedx
> Tests whether the given optional value is defined, i.e. has a non-`None` value.

(In CWL we would represent `None` as `null` in the context of a CWL Expression or CWL Reference)

In Javascript, referencing a condition in a if-then-else statement or by using the ternary operator `?` evaluates if the condition (the element before the `?`) is [truthy](https://developer.mozilla.org/en-US/docs/Glossary/Truthy) or [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy).

Example: the `inputsR2` variable could have the boolean value `False`, but `defined(inputsR2)` should return `true`.

Some we searching reveals that to determine if a variable has the value `null` in Javascript we can use the strict equality operator: `variable_name === null` (`true` if the variable is`null` (not defined), `false` otherwise). As the WDL operator `defined` has the reverse meaning (`true` if the variable is defined, `false` otherwise) we can use the strict inequality operator: `variable_name !== null`.

Here is some code I ran at https://replit.com/languages/javascript that explores this idea

``` javascript
var inputR2_a = null;
var inputR2_b = "";
var inputR2_c = false;
var inputR2_d = {"type": "File", "path": "local/path/reads.fq"};

console.log(inputR2_a ? 'inputsR2_a is truthy.' : 'inputR2_a is falsy.');
console.log(inputR2_b ? 'inputsR2_b is truthy.' : 'inputR2_b is falsy.');
console.log(inputR2_c ? 'inputsR2_c is truthy.' : 'inputR2_c is falsy.');
console.log(inputR2_d ? 'inputsR2_d is truthy.' : 'inputR2_d is falsy.');
console.log();
console.log(inputR2_a === null ? 'inputsR2_a is not defined (is null).' : 'inputR2_a is defined (is not null)');
console.log(inputR2_b === null ? 'inputsR2_b is not defined (is null).' : 'inputR2_b is defined (is not null)');
console.log(inputR2_c === null ? 'inputsR2_c is not defined (is null).' : 'inputR2_c is defined (is not null)');
console.log(inputR2_d === null ? 'inputsR2_d is not defined (is null).' : 'inputR2_d is defined (is not null)');
```

```
inputR2_a is falsy.
inputR2_b is falsy.
inputR2_c is falsy.
inputsR2_d is truthy.

inputsR2_a is not defined (is null).
inputR2_b is defined (is not null)
inputR2_c is defined (is not null)
inputR2_d is defined (is not null)
```

dinithiW: We changed the default value of optional  inputs to ''. So it won't be necessary to check whether the value is null right? All optional inputs will look like this if not defined ```var inputR2_b = "";```

mr-c: Okay, then modify the javascript to check for equality with `""` (not defined), otherwise they are defined

mr-c (approved):

PR #48, opened 2021-08-03
Modify Wdl output

dinithiW:
New test case has been added.
Modified code for wdl outputs

### DIFF
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -194,9 +194,9 @@ def get_output(expression: str, input_names: List[str]) -> str:
         output_value = output_value.replace('"', "")
 
     elif '"' not in expression:
-        if expression.replace('"', "") in input_names:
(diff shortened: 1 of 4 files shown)


wdl2cwl/main.py
```diff
@@ -197,6 +197,10 @@ def get_output(expression: str, input_names: List[str]) -> str:
         if expression.replace('"', "") in input_names:
             output_value = "$(inputs." + expression + ")"
         output_value = output_value.replace('"', "")
```
mr-c, line 199: Why the `replace()` in both of these lines? We already know that `"` is not in `expression` from the ` elif '"' not in expression:`

dinithiW: Yes. Not necessary. I do it out of habit :D Modified the code

mr-c (approved):

PR #54, opened 2021-08-09
Handle Meta

dinithiW:
Issues a warning when a meta section is encountered in a WDL document.
Added new test case

### DIFF
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -426,6 +426,9 @@ def convert(workflow: str) -> str:
     if ast.task_parameter_meta_check:
         print("----WARNING: SKIPPING PARAMETER_META----")
 
+    if ast.task_meta_check:
+        print("----WARNING: SKIPPING META----")
+
     if len(ast.task_variables) > 0:
         for a in ast.task_variables:
             print("----WARNING: SKIPPING VARIABLE " + str(a[1]) + "----")
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -421,17 +421,19 @@ def convert(workflow: str) -> str:
 
     for i in ast.task_runtime:
         if i not in runtime_requirements:
-            print("----WARNING: SKIPPING REQUIREMENT " + i + "----")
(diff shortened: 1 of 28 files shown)


wdl2cwl/main.py
```diff
@@ -426,6 +426,9 @@ def convert(workflow: str) -> str:
     if ast.task_parameter_meta_check:
         print("----WARNING: SKIPPING PARAMETER_META----")
 
+    if ast.task_meta_check:
+        print("----WARNING: SKIPPING META----")
+
```
mr-c, line 431: Is there a test that confirms this warning is printed?

mr-c (changes requested):
Looks good! Needs two specific tests to confirm that warning is printed when it should be; and not printed when it shouldn't

wdl2cwl/main.py
dinithiW: I'm not sure why UmiCorrection.wdl won't work for this. It has a meta section.

mr-c: Use https://docs.pytest.org/en/6.2.x/reference.html#std-fixture-capsys to check that the warning is printed

dinithiW:
```
wdl2cwl/tests/test_cwl.py: note: In function "test_meta":
  wdl2cwl/tests/test_cwl.py:13: error: Missing type parameters for generic type
  "CaptureFixture"  [type-arg]
      def test_meta(capsys: pytest.CaptureFixture) -> None:
      ^
  Found 1 error in 1 file (checked 4 source files)
```
Is the type incorrect?

mr-c:
> ```
> wdl2cwl/tests/test_cwl.py: note: In function "test_meta":
>   wdl2cwl/tests/test_cwl.py:13: error: Missing type parameters for generic type
>   "CaptureFixture"  [type-arg]
>       def test_meta(capsys: pytest.CaptureFixture) -> None:
>       ^
>   Found 1 error in 1 file (checked 4 source files)
> ```
> Is the type incorrect?

Try `def test_meta(capsys: pytest.CaptureFixture[str]) -> None:`

mr-c (approved):

PR #55, opened 2021-08-09
Handle expression placeholders in command section

dinithiW:
Handles expression placeholders with ```${}``` syntax in the command section of the WDL document.
New test case has been added.


wdl2cwl/tests/wdl_files/UmiCorrection.wdl
```diff
@@ -0,0 +1,80 @@
+version 1.0
+
+task CorrectUMItools {
```
mr-c, line 3: Where is this file from?

wdl2cwl/main.py
```diff
@@ -87,7 +87,9 @@ def get_command(
     while index < len(command):
 
         # if you have ~{
-        if command[index] == "~" and command[index + 1] == "{":
+        if (command[index] == "~" and command[index + 1] == "{") or (
+            command[index] == "$" and command[index + 1] == "{"
```
mr-c, line 91: Only replace `${…}` style if the command block is in the `command { ... }`style (and never if the command block is in the `command <<< >>>` style): https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#expression-placeholders

Also, don't replace if the identifier is not the name of any of the inputs.

mr-c (changes requested):

dinithiW:
Modified code as requested

PR #78, opened 2021-08-24
Handle optional array inputs

dinithiW:
Handles optional array inputs. Added new test case.

- ~~Fix the "" defaults for non-string types~~ Moved to #80 
- [x] The arrays of files should get the .path attribute
- [x] Apply the changes from https://github.com/biowdl/tasks/pull/291/files#diff-c76c01f3ca967cdb9c157a75e7fb1a08d0037543b455c2107398601a2f526ebfR1081 to fix the file co-location error in the original WDL
- [ ] There is an extra prefix for the arrays. The separator should only be added if the current element is not the last one in the array.

### DIFF
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -322,15 +322,12 @@ def get_input(
             )
             input_name = i[1]
 
-            if "?" not in i[0]:
-                inputs.append(
-                    cwl.CommandInputParameter(
-                        id=input_name,
-                        type=[
-                            cwl.CommandInputArraySchema(items=input_type, type="array")
-                        ],
-                    )
+            inputs.append(
+                cwl.CommandInputParameter(
+                    id=input_name,
+                    type=[cwl.CommandInputArraySchema(items=input_type, type="array")],
                 )
+            )
 
         else:
             input_type = (
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -183,9 +183,10 @@ def get_command(
                 if input_name in input_names:
                     append_str = (
                         '${\nvar text = "";\n'
-                        + 'for(var i=0;i<inputs["'
+                        + 'var arr_length = inputs["'
                         + input_name
-                        + '"].length;i++) \n'
+                        + '"].length;\n'
+                        + "for(var i=0;i<arr_length-1;i++) \n"
                         + '  text+= inputs["'
                         + input_name
                         + '"][i]'
@@ -193,6 +194,11 @@ def get_command(
                         + '+"'
                         + separator
                         + '";\n'
+                        + 'text+= inputs["'
+                        + input_name
+                        + '"][arr_length-1]'
+                        + temp_append_str
+                        + ";\n"
                         + "return text;\n"
                         + "}"
                     )
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -173,6 +173,12 @@ def get_command(
                 split_str = sub_str.split(sub_str[4])
                 separator = split_str[1]
                 input_name = split_str[2]
+                index = input_names.index(input_name)
+                data_type = input_types[index]
+
+                temp_append_str = ""
+                if "Array[File]" in data_type:
+                    temp_append_str = ".path"
 
                 if input_name in input_names:
                     append_str = (
@@ -182,7 +188,9 @@ def get_command(
                         + '"].length;i++) \n'
                         + '  text+= inputs["'
                         + input_name
-                        + '"][i]+"'
+                        + '"][i]'
+                        + temp_append_str
+                        + '+"'
                         + separator
                         + '";\n'
                         + "return text;\n"
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -147,11 +147,18 @@ def get_command(
 
                 elif "defined(" in input_name:
                     sub_str = input_name[input_name.find("(") + 1 : -1]
+                    index = input_names.index(sub_str)
+                    data_type = input_types[index]
+                    check_str = ""
+                    if "Array" in data_type:
+                        check_str=".length === 0 "
+                    else:
+                        check_str=' === "" '
                     append_str = (
                         '$(inputs["'
                         + sub_str
-                        + '"] === '
-                        + '"" ? '
+                        + '"]' 
+                        + check_str +'? '
                         + '"'
                         + false_value
                         + '"'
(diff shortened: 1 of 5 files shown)


wdl2cwl/main.py
```diff
@@ -331,6 +331,16 @@ def get_input(
...
                     )
                 )
+            else:
+                inputs.append(
+                    cwl.CommandInputParameter(
+                        id=input_name,
+                        type=[
+                            cwl.CommandInputArraySchema(items=input_type, type="array")
+                        ],
+                        default="",
```
mr-c, line 341: `[]` would be a better default, as an empty string is doesn't match the type.

`defined` would also need to work differently for arrays, testing for length greater than zero instead of (in)equality with and empty string.

mr-c (changes requested):

mr-c:
Starting to get a little complicated! Can you try to find test data for this one, preferably using at least one of the optional arrays?

dinithiW:
> Starting to get a little complicated! Can you try to find test data for this one, preferably using at least one of the optional arrays?

Any idea what kinds of files can go to intervalList?

dinithiW:
Changed default value and JS expression for defined function

mr-c:
> > Starting to get a little complicated! Can you try to find test data for this one, preferably using at least one of the optional arrays?
> 
> Any idea what kinds of files can go to intervalList?

https://gatk.broadinstitute.org/hc/en-us/articles/360035531852-Intervals-and-interval-lists gives some hints

mr-c:
>     1. What should the default value be? I think we had `null` as the default value first, but changed to `""` because we didn't want null to be passed into the command.

Yeah, that made sense at the beginning of this project. Now we see that we need a new approach, so it is time to improve it.

After reading https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#concatenation-of-optional-values I think that we may need to re-do how we handle optional WDL inputs. 

It seems the safer thing to do is
1. Not define a default value for optional inputs (unless the WDL document has one)

and

2. For each WDL expression placeholder `~{…}` containing references to one or more **optional** inputs
  a. replace it with CWL expression that returns an empty string if **any** of the optional inputs referenced in the WDL expression are missing (`=== null`) and otherwise returns the result of the expression itself.

Do you think this plan will work?

If you want, you can make a new issue to fix the optional WDL inputs, and then this PR can be finished by removing  temporarily removing the optional inputs from `gatk_1.wdl` (after Apply the changes from https://github.com/biowdl/tasks/pull/291/files#diff-c76c01f3ca967cdb9c157a75e7fb1a08d0037543b455c2107398601a2f526ebfR1081 to fix the file co-location error in the original WDL)

dinithiW:
> Do you think this plan will work?
> 
> If you want, you can make a new issue to fix the optional WDL inputs, and then this PR can be finished by removing temporarily removing the optional inputs from `gatk_1.wdl` (after Apply the changes from https://github.com/biowdl/tasks/pull/291/files#diff-c76c01f3ca967cdb9c157a75e7fb1a08d0037543b455c2107398601a2f526ebfR1081 to fix the file co-location error in the original WDL)

Ok. I will create a new PR to handle the optional inputs. I have updated gatk_1.wdl after removing the optional inputs from the inputs section and the command.

wdl2cwl/tests/wdl_files/gatk_1.wdl
```diff
@@ -34,15 +34,15 @@ task HaplotypeCaller {
         #String emitRefConfidence = if gvcf then "GVCF" else "NONE"
         Boolean dontUseSoftClippedBases = false
 
-        Array[File]+? intervalList
-        Array[File]+? excludeIntervalList
```
mr-c: Let's keep these two as they are the topic of this PR (optional array inputs) and the conversion worked.

dinithiW: Ok. If I apply all the changes from https://github.com/biowdl/tasks/pull/291/files#diff-c76c01f3ca967cdb9c157a75e7fb1a08d0037543b455c2107398601a2f526ebfR1081 the resultant cwl is not accurate. Shall I keep it as it is or remove the changes?

mr-c: Can you show me what is not accurate about the resulting CWL?

wdl2cwl/tests/wdl_files/gatk_1.wdl
```diff
@@ -0,0 +1,106 @@
...
+        Int timeMinutes = 400 # This will likely be used with intervals, as such size based estimation can't be used.
+        String dockerImage = "quay.io/biocontainers/gatk4:4.1.8.0--py38h37ae868_0"
+    }
+
+    command {
+        set -e
+        mkdir -p "$(dirname ~{outputPath})"
+        gatk --java-options '-Xmx~{javaXmxMb}M -XX:ParallelGCThreads=1' \
+        HaplotypeCaller \
+        -R wd2/~{basename(referenceFasta)} \
```
mr-c, line 59: Good that you copied in the `wd2/` change, but this doesn't work without the `ln -s` changes that go with it https://github.com/biowdl/tasks/pull/291/files#diff-c76c01f3ca967cdb9c157a75e7fb1a08d0037543b455c2107398601a2f526ebfR1084

wdl2cwl/tests/cwl_files/gatk_1.cwl
```diff
@@ -0,0 +1,72 @@
...
+            set -e
+            mkdir -p "\$(dirname $(inputs.outputPath))"
+            gatk --java-options '-Xmx$(inputs.javaXmxMb)M -XX:ParallelGCThreads=1' \
+            HaplotypeCaller \
+            -R wd2/$(inputs.basename(referenceFasta)) \
+            -O $(inputs.outputPath) \
+            -I ${
+            var text = "";
+            for(var i=0;i<inputs["inputBams"].length;i++) 
+              text+= inputs["inputBams"][i].path+" -I ";
```
mr-c, line 61: This is still not working.

There is an extra prefix for the arrays. The separator should only be added if the current element is not the last one in the array.

wdl2cwl/tests/wdl_files/gatk_1.wdl
dinithiW: For example ```for FILE in $(inputs.sep(" ",inputBams)); do ln -s $FILE wd/\$(inputBams $FILE) ; done```

wdl2cwl/tests/wdl_files/gatk_1.wdl
dinithiW: Yes. I only saw those changes later :/

wdl2cwl/tests/cwl_files/gatk_1.cwl
dinithiW: I modified the code to address this

PR #83, opened 2021-09-01
Handle optional inputs

### LINKED ISSUE #80: re-do optional inputs default values
>     1. What should the default value be? I think we had `null` as the default value first, but changed to `""` because we didn't want null to be passed into the command.

Yeah, that made sense at the beginning of this project. Now we see that we need a new approach, so it is time to improve it.

After reading https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#concatenation-of-optional-values I think that we may need to re-do how we handle optional WDL inputs. 

It seems the safer thing to do is
1. Not define a default value for optional inputs (unless the WDL document has one)

and

2. For each WDL expression placeholder `~{…}` containing references to one or more **optional** inputs
  a. replace it with CWL expression that returns an empty string if **any** of the optional inputs referenced in the WDL expression are missing (`=== null`) and otherwise returns the result of the expression itself.

_Originally posted by @mr-c in https://github.com/common-workflow-lab/wdl-cwl-translator/issues/78#issuecomment-907755829_

dinithiW:
Fixes #80 

Removed default value for optional inputs.
Expressions are written for optional inputs without a default value.
Updated the test files


wdl2cwl/tests/cwl_files/deepvariant.cwl
```diff
@@ -103,12 +96,12 @@ requirements:
             --reads $(inputs.inputBam.path) \
             --model_type $(inputs.modelType) \
             --output_vcf $(inputs.outputVcf) \
-            --output_gvcf $(inputs.outputGVcf) \
-            --customized_model $(inputs.customizedModel) \
-            --num_shards $(inputs.numShards) \
-            --regions $(inputs.regions) \
-            --sample_name $(inputs.sampleName) \
-            --postprocess_variants_extra_args $(inputs.postprocessVariantsExtraArgs) \
+            --output_gvcf $(inputs["outputGVcf"] === null ? "" : inputs["outputGVcf"]) \
```
mr-c, line 99: The original WDL expression is `~{"--output_gvcf " + outputGVcf} \` so the CWL version should be
```suggestion
            $(inputs["outputGVcf"] === null ? "" : "--output_gvcf " + inputs["outputGVcf"]) \
```

This is the tricky part.

If a WDL expression has even a single reference to an optional input (even if there are other, non-optional references) and that input is missing (`=== null`) then the entire WDL expression should result in an empty string.

mr-c (changes requested):

wdl2cwl/tests/cwl_files/deepvariant.cwl
dinithiW: :/ That requires some work. What will happen if there's two variables inside the expression ```~{}``` where only one is optional and the other is a required variable. Should the entire expression still evaluate to `null`?

mr-c: > What will happen if there's two variables inside the expression ~{} where only one is optional and the other is a required variable. Should the entire expression still evaluate to null?

Yes. Tricky!

wdl2cwl/tests/cwl_files/TrimAdapters.cwl
```diff
@@ -44,7 +43,7 @@ requirements:
             fastq-mcf \
                -C 200000 $(inputs.adapter_list.path) \
                $(inputs.fastq1.path) \
-               $(inputs.fastq2) \
+               $(inputs["fastq2"] === null ? "" : "inputs["fastq2"].path") \
```
mr-c, line 46:
```suggestion
               $(inputs["fastq2"] === null ? "" : inputs["fastq2"].path) \
```
Can't have quotation marks here

wdl2cwl/tests/cwl_files/bowtie_1.cwl
```diff
@@ -68,14 +64,14 @@ requirements:
             bowtie \
             -q \
             --sam \
-            --seedmms $(inputs.seedmms) \
-            --seedlen $(inputs.seedlen) \
-            -k $(inputs.k) \
+            $(inputs["seedmms"] === null ? "" : "--seedmms inputs["seedmms"]") \
+            $(inputs["seedlen"] === null ? "" : "--seedlen inputs["seedlen"]") \
+            $(inputs["k"] === null ? "" : "-k inputs["k"]") \
```
mr-c, line 69:
```suggestion
            $(inputs["seedmms"] === null ? "" : "--seedmms " + inputs["seedmms"]) \
            $(inputs["seedlen"] === null ? "" : "--seedlen " + inputs["seedlen"]) \
            $(inputs["k"] === null ? "" : "-k "+ inputs["k"]) \
```

Another quotation mark issue

mr-c (changes requested):
Looking much better!

Some quotation marks are in the wrong place (or shouldn't be there at all)

Did you run the generated CWL to make sure the shell script is valid?

Exciting!

dinithiW:
> Did you run the generated CWL to make sure the shell script is valid?

No I couldn't run them. I'll run them to look for any errors.

dinithiW:
Checked the validity of the CWL files
Modified code as required

wdl2cwl/tests/cwl_files/hisat2_1.cwl
```diff
@@ -66,8 +63,8 @@ requirements:
             hisat2 \
             -p $(inputs.threads) \
             -x $(inputs.indexFiles[0]) \
-            $(inputs["inputR2"] === "" ? "-U": "-1") $(inputs.inputR1.path) \
-            -2$(inputs.inputR2) \
+            $(inputs["inputR2"].path === null ? "-U": "-1") $(inputs.inputR1.path) \
+            $(inputs["inputR2"].path === null ? "" : "-2" + inputs["inputR2"].path ) \
```
mr-c, line 67:
```suggestion
            $(inputs["inputR2"] === null ? "-U": "-1") $(inputs.inputR1.path) \
            $(inputs["inputR2"] === null ? "" : "-2" + inputs["inputR2"].path ) \
```
If `inputR2` is not provided then `inputs["inputR2"]` will be `null` and the reference `inputs["inputR2"].path` will make an error

mr-c (changes requested):
Very close to being done!

dinithiW:
Merged the changes and modified code as requested

mr-c (approved):

Issue #81, opened 2021-08-30
WDL read_string


mr-c:
https://github.com/broadinstitute/warp/blob/cec97750e3819fd88ba382534aaede8e05ec52df/tests/skylab/optimus/pr/ValidateOptimus.wdl#L139

`id = "result", type = "string", outputBinding=cwl.CommandOutputBinding(glob="result.txt", load_contents=True, outputEval= "$(self.contents.replace(/[\r\n]+$/, '')")`

Th3nn3ss:
Could you add a description? so I can understand what the issue is.

mr-c:
Thanks for the reminder @Th3nn3ss ; I've added more detail to [the issue](https://github.com/common-workflow-lab/wdl-cwl-translator/issues/81#issue-982958902)

Th3nn3ss:
Thanks for the detail to the issue. I am understanding the code base better. From the equivalent code you provided, the argument "load_contents=True" for the CommandOutputBinding function is throwing an unexpected keyword argument error.

PR #88, opened 2021-10-11
 issue #81: WDL read_string

Th3nn3ss:
Handle read_string function when it's present in output section

### DIFF
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -659,6 +659,15 @@ def convert(workflow: str) -> str:
         output_name = i[1]
         output_glob = get_output(i[2], input_names)
 
+        if "String" in i[0]:
+            outputs.append(
+                cwl.CommandOutputParameter(
+                    id=output_name, type="string", 
+                    outputBinding=cwl.CommandOutputBinding(glob="result.txt", 
+                    outputEval="$(self.contents.replace(/[\r\n]+$/, '')"))
+            )
+            continue
+
         if "Array" in i[0]:
             # output_type = ""
             temp_type = wdl_type[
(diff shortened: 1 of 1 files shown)


mr-c (changes requested):
Thank you @Th3nn3ss 

Can you include a test case? There is an example mentioned at https://github.com/common-workflow-lab/wdl-cwl-translator/issues/81 so you can skip directly to step 2 of https://github.com/common-workflow-lab/wdl-cwl-translator#adding-test-cases

wdl2cwl/main.py
```diff
@@ -659,6 +659,19 @@ def convert(workflow: str) -> str:
         output_name = i[1]
         output_glob = get_output(i[2], input_names)
 
+        if "String" in i[0]:
```
dinithiW, line 662: This conversion is not required everytime a String output is encountered. The conversion is needed for String outputs where the content on the right side of the `=` sign in the WDL task (`i[1]` in this case) contains `read_string`. So the `if` condition here should be modified.

wdl2cwl/tests/test_cwl.py
```diff
@@ -47,6 +47,22 @@ def test_meta(capsys: pytest.CaptureFixture[str]) -> None:
         ("wdl_files/transcriptclean_1.wdl", "cwl_files/transcriptclean_1.cwl"),
         ("wdl_files/gatk_1.wdl", "cwl_files/gatk_1.cwl"),
         ("wdl_files/deepvariant.wdl", "cwl_files/deepvariant.cwl"),
+        (
+            "wdl2cwl/tests/wdl_files/validateOptimus_1.wdl",
+            "wdl2cwl/tests/cwl_files/validateOptimus_1.cwl",
+        )(
```
mr-c, line 53: There should be commas between the tuples in this section

Th3nn3ss:
`make test` is returning this error `make: *** No rule to make target 'wdl2cwl/cwl_files/*.cwl', needed by 'cwltest'.  Stop.
`

PR #101, opened 2021-11-01
Translate runtime cpu into coresMin even when string provided

### LINKED ISSUE #98: WDL runtime requirement cpu not being parsed correctly
Hi,

For #97 , I used the warp project [BuildIndices](https://github.com/broadinstitute/warp/tree/develop/pipelines/skylab/build_indices) pipeline. That pull request handles structs, so the rest of the pipeline is also converted into CWL.

When I tried to run the generated CWL workflow, it failed with:

```
...
test.cwl:54:5: object id `test.cwl#BuildIntervalList/genome_fa` previously defined
ERROR Tool definition failed validation:
test.cwl:1:1:     Object `test.cwl#BuildIntervalList` is not valid because
                  tried `CommandLineTool` but
test.cwl:197:1:     the `requirements` field is not valid because
                      tried array of <InlineJavascriptRequirement or SchemaDefRequirement or
                      LoadListingRequirement or DockerRequirement or SoftwareRequirement or
                      InitialWorkDirRequirement or EnvVarRequirement or ShellCommandRequirement or
                      ResourceRequirement or WorkReuse or NetworkAccess or InplaceUpdateRequirement or
                      ToolTimeLimit or SubworkflowFeatureRequirement or ScatterFeatureRequirement or
                      MultipleInputFeatureRequirement or StepInputExpressionRequirement> but
test.cwl:229:5:         item is invalid because
test.cwl:230:5:           the `coresMin` field is not valid because
                            - tried long but
                              the value `''` is not long
                            - tried float but
                              the value `''` is not float or double
                            - tried Expression but
                              value `` does not contain an expression in the form $() or ${}
```

Looking at the CWL requirements, `coresMin` was empty.

```
  - class: ResourceRequirement
    ramMin: 8192
  - class: ResourceRequirement
    coresMin: ''
```

Debugging it in the IDE, it looks like the logic is ignoring any values that contain `"`. But I think it forgot to initialize `cpu` with a value?

image

In the screenshot above, see in the bottom right corner, the watched value `ast.task_runtime["cpu"]`, that contains `'"8"'` (number eight wrapped between double quotes). Then see the `if` statement (collapsed in the IDE for simplify it) where it only sets a value for `cpu` if the value does not contain quotes (it doesn't have an `else` statement).

So the final value is an empty String in the CWL file, resulting in the error I had. I think a better solution would be to parse the `8` integer from the quoted text?

kinow:
Closes #98

### DIFF
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -666,23 +666,23 @@ def convert(workflow: str) -> str:
     if "cpu" in ast.task_runtime:
 
         cpu: Union[str, int] = ""
-        if '"' not in ast.task_runtime["cpu"]:
-            if ast.task_runtime["cpu"] in input_names:
-                cpu = f"$({inputs(ast.task_runtime['cpu'])})"
-            elif ast.task_runtime["cpu"].isnumeric():
-                cpu = int(ast.task_runtime["cpu"])
-            elif "+" in ast.task_runtime["cpu"]:
-                temp = ast.task_runtime["cpu"].split("+")
-                append_str = "$("
-                for i in range(0, len(temp)):
-                    if temp[i] in input_names:
-                        append_str += inputs(temp[i])
-                    elif temp[i].isnumeric():
-                        append_str += temp[i]
-                    if i != len(temp) - 1:
-                        append_str += " + "
-                append_str += ")"
-                cpu = append_str
+        cpu_value = ast.task_runtime["cpu"].strip("'\" ")
+        if cpu_value in input_names:
+            cpu = f"$({inputs(cpu_value)})"
+        elif cpu_value.isnumeric():
+            cpu = int(cpu_value)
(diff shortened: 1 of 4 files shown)


wdl2cwl/main.py
```diff
@@ -666,23 +666,23 @@ def convert(workflow: str) -> str:
...
-                for i in range(0, len(temp)):
-                    if temp[i] in input_names:
-                        append_str += inputs(temp[i])
-                    elif temp[i].isnumeric():
-                        append_str += temp[i]
-                    if i != len(temp) - 1:
-                        append_str += " + "
-                append_str += ")"
-                cpu = append_str
+        cpu_value = ast.task_runtime["cpu"].strip("'\" ")
```
kinow, line 669: The suggested fix is implemented here :point_up: the rest of the code is pretty much the same.

Issue #59, opened 2021-08-12
WDL function select_first() is not handled


dinithiW:
Currently, the WDL function select_first() is not translated by the convertor. (https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#x-select_firstarrayx)

A WDL workflow containing select_first() function can be found here. (https://github.com/biowdl/tasks/blob/develop/bwa.wdl#L82)

Th3nn3ss:
Is there a javascript equivalent for the WDL `select_first()` function?

Th3nn3ss:
could I use the `find` function? 

Say, in the case of the example provided in (https://github.com/biowdl/tasks/blob/develop/bwa.wdl#L82):
`memory: "~{select_first([memoryGb, estimatedMemoryGb])}G"` would be converted to 
`memory: "${[memoryGb, estimatedMemoryGb].find(memory => moemory != None)}G"`

kinow:
Hi @Th3nn3ss 

I think in JavaScript it would have to use strings `['memoryGb', 'estimatedMemoryGb']`. And the Python `None` does not exist in JavaScript. So probably treat it as `null` or `undefined` (or check if the None mentioned in the [WDL spec](https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#x-select_firstarrayx) should be treated differently in the wdl-cwl-translator.)

Bruno

Th3nn3ss:
> Hi @Th3nn3ss
> 
> I think in JavaScript it would have to use strings `['memoryGb', 'estimatedMemoryGb']`, unless these two are variales declared previously. And the Python `None` does not exist in JavaScript, so maybe test for `null` and `undefined` (or check if the None mentioned in the [WDL spec](https://github.com/openwdl/wdl/blob/main/versions/development/SPEC.md#x-select_firstarrayx) should be treated differently in the wdl-cwl-translator.)
> 
> Bruno

OK. 
The WDL documentation actually says it returns the first non-`None` value of an `Array` with optional values. So I have to find a way for a `None` to be treated as a Javascript `null' so it can return the appropriate error message when the expression isn't successful. Errors like `error! array contains only None values`

mr-c:
> > Hi @Th3nn3ss
> > 
[...]

PR #102, opened 2021-11-01
support select_first function for non-None values that were declared as inputs

Th3nn3ss:
* The select_first function selects the first non-`None` value from an array of values. 
* This fix uses refactors the `get_ram_min_js` function to accept a list of inputs and return the first non-`null` value of a valid javascript array of the inputs variables.
* The expression contains a `for loop` that returns the first non-`null` element and also throws an error when only null values are found or the array is empty.
* Undeclared inputs are not handled as the wdl-cwl translator does not support such inputs yet.

Fixes: #59


wdl2cwl/main.py
```diff
@@ -632,6 +632,26 @@ def convert(workflow: str) -> str:
                     unit = ast.task_runtime["memory"][temp + 1 : -1].strip()
                     if input_name in input_names:
                         ram_min = get_ram_min_js(input_name, unit)
+            elif "~{select_first(" in ast.task_runtime["memory"]:
```
mr-c, line 635: for processing `select_first` in the context of `task_runtime["memory"]`, please send the results to [`get_ram_min_js()`](https://github.com/common-workflow-lab/wdl-cwl-translator/blob/main/wdl2cwl/main.py#L67)

mr-c (changes requested):
Did you run the resulting CWL yet?

wdl2cwl/main.py
Th3nn3ss: I should treat the `select_first` function as a list of inputs and send the first valid input to `get_ram_min_js()`?

wdl2cwl/tests/wdl_files/bwa.wdl
```diff
@@ -0,0 +1,112 @@
...
+        String dockerImage = "quay.io/biocontainers/mulled-v2-ad317f19f5881324e963f6a6d464d696a2825ab6:c59b7a73c87a9fe81737d5d628e10a3b5807f453-0"
+    }
+
+    # Samtools sort may block the pipe while it is writing data to disk.
+    # This can lead to cpu underutilization.
+    # 1 thread if threads is 1. For 2-4 threads 2 sort threads. 3 sort threads for 5-8 threads.
+    # Int estimatedSortThreads = if threads == 1 then 1 else 1 + ceil(threads / 4.0)
+    # Int totalSortThreads = 2 # select_first([sortThreads, estimatedSortThreads])
+    # BWA needs slightly more memory than the size of the index files (~10%). Add a margin for safety here.
+    # Int estimatedMemoryGb = 1 + ceil(size(bwaIndex.indexFiles, "G") * 1.2) + sortMemoryPerThreadGb * totalSortThreads
```
mr-c, line 50: There is not point in testing `select_first` with only one input. So please turn `estimatedMemoryGb` into an input of type `Int` like `memoryGb`

wdl2cwl/main.py
mr-c: no; you should send the javascript snippet to `get_ram_min_js` and modify that function to incorporate it

mr-c (changes requested):
should send the javascript snippet to `get_ram_min_js()` and modify that function to incorporate it; test the result to confirm the fix

Th3nn3ss:
> should send the javascript snippet to `get_ram_min_js()` and modify that function to incorporate it; test the result to confirm the fix

Can I get more clarification on this? I don't fully understand wish javascript snippet  am passing to `get_ram_min_js()`.

mr-c:
> > should send the javascript snippet to `get_ram_min_js()` and modify that function to incorporate it; test the result to confirm the fix
> 
> Can I get more clarification on this? I don't fully understand wish javascript snippet am passing to `get_ram_min_js()`.

from your code: https://github.com/common-workflow-lab/wdl-cwl-translator/pull/102/commits/ee4eb916dc4dbdc3251d6eb50bbe1902ae231b32#diff-b5a65c2befbbd0b6453805272729b622f41b2be683f8258432e08f4370332265R652

Except that `Array.find` is not part of ECMAScript 5.1, so you'll have to implement your own function.

Remember, according to https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#x-select_firstarrayx your ECMAScript version should throw an exception if the array is empty or if the array only contains `null` values.

mr-c:
It might be easier to
1) make a copy of `get_ram_min_js` and edit it to be convenient and produce working CWL expressions
2) then compare that new function to the other uses of `get_ram_min_js`
3) then figure out how to refactor `get_ram_min_js` so that it supports this new use case (needing to evaluate some ECMAScript code in the same function that does dynamic unit conversion) with its existing usage.

mr-c (changes requested):
Please undelete `.gitignore'

You've completed step 1 from https://github.com/common-workflow-lab/wdl-cwl-translator/pull/102#issuecomment-962430578 ; now please consider step 2 and 3.

Specifically, the `select_first` javascript logic should not go inside `get_ram_min_js`.

Maybe split the implementation of `get_ram_min_js` into two:
1. `get_ram_min_js_from_reference()` that takes a plain reference and an optional unit (essentially the same as the [existing `get_ram_min_js`](https://github.com/common-workflow-lab/wdl-cwl-translator/blob/7cd094a27f4e74ba5a3d67dc2aed0fece6766d99/wdl2cwl/main.py#L80))
2. `get_ram_min_js_from_function()`  that accepts a Javascript "preamble" which must result in the Javascript variable `ram_min` being set.

The two functions would both append the same Python constant that contains the epilogue: code to convert the units and `return parseInt(memory);` (from https://github.com/common-workflow-lab/wdl-cwl-translator/blob/7cd094a27f4e74ba5a3d67dc2aed0fece6766d99/wdl2cwl/main.py#L95 )

Th3nn3ss:
HI @mr-c , The CI tests keeps reporting this 
```
  Traceback:
  /opt/hostedtoolcache/Python/3.7.12/x64/lib/python3.7/importlib/__init__.py:127: in import_module
      return _bootstrap._gcd_import(name[level:], package, level)
  wdl2cwl/tests/test_cwl.py:6: in <module>
      from .. import main as wdl
  wdl2cwl/main.py:13: in <module>
      import cwl_utils.parser_v1_2 as cwl
  E   ModuleNotFoundError: No module named 'cwl_utils.parser_v1_2'
```

mr-c:
> HI @mr-c , The CI tests keeps reporting this

Thanks. I just pushed a fix for that.

Th3nn3ss:
Forgive my confusion. I am not sure what part of the requested change am yet to implement. I already split the function and added the epilogue js string.

wdl2cwl/main.py
```diff
@@ -107,6 +107,34 @@ def get_ram_min_js(ram_min: str, unit: str) -> str:
...
+        + 'else if(unit==="MiB") memory = value;\n'
+        + 'else if(unit==="GiB") memory = value*1024;\n'
+        + 'else if(unit==="TiB") memory = value*1024*1024;\n'
+        + 'else if(unit==="B") memory = value/(1024*1024);\n'
+        + 'else if(unit==="KB" || unit==="K") memory = (value*1000)/(1024*1024);\n'
+        + 'else if(unit==="MB" || unit==="M") memory = (value*(1000*1000))/(1024*1024);\n'
+        + 'else if(unit==="GB" || unit==="G") memory = (value*(1000*1000*1000))/(1024*1024);\n'
+        + 'else if(unit==="TB" || unit==="T") memory = (value*(1000*1000*1000*1000))/(1024*1024);\n'
+        + "return parseInt(memory);\n}"
+    )
```
mr-c, line 133: Please refactor this function and `get_ram_min_js_from_reference` so there is less code duplication

mr-c (changes requested):

mr-c:
Closing as this needs to be redone from scratch given the refactoring. Hopefully it will be easier the 2nd time!

PR #106, opened 2021-12-08
first draft with miniwdl

Th3nn3ss:
This is just a draft of the Object Oriented Approach with miniwdl.


Th3nn3ss:
I can't seem to get the instances to return true for when the inputs match specific miniwdl WDL type objects like WDL.Type.Int, WDL.Type.Array

Th3nn3ss:
I tried it exactly as was described in the [miniwdl documentation] (https://miniwdl.readthedocs.io/en/latest/WDL.html#module-WDL.Type)

PR #107, opened 2021-12-12
Miniwdl oop

Th3nn3ss:
Draft to use miniwdl to implement an object oriented version of the wdl2cwl converter.

1. Convert a single wdl file to cwl by hand without loops, if conditions or classes.
    - Handled inputs
    - Handled outputs
    - Handled command section
    - Handled runtime section

### DIFF
--- a/wdl2cwl/main_oop.py
+++ b/wdl2cwl/main_oop.py
@@ -279,6 +279,8 @@ if __name__ == '__main__':
 
     wdl_command_part_1 = command.parts[0]
     command_1 = textwrap.dedent(wdl_command_part_1)
+    command_1_splitted_p1, command_1_splitted_p2 = command_1.split("$")
+    command_1 = command_1_splitted_p1 + "\\$" + command_1_splitted_p2
     cwl_command_str += command_1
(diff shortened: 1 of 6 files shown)


wdl2cwl/tests/oop_cwl_files/bowtie_1.cwl
```diff
@@ -0,0 +1,60 @@
+class: CommandLineTool
+id: Bowtie
```
mr-c, line 2: Does this file match the `main.py` output?

Th3nn3ss: Yes. But the requirements is incomplete.

Am still exploring the command section from the WDL file.

wdl2cwl/main_oop.py
```diff
@@ -0,0 +1,584 @@
+import os
+from typing import List, Union, Optional, Callable
+import WDL
```
mr-c, line 3:
```suggestion
import WDL  # type: ignore[import]
```

kinow (approved):
:drum: great job @Th3nn3ss !+1

PR #112, opened 2021-12-16
refactor miniwdl proof of concept using some functions

### LINKED ISSUE #109: Design classes and relationships for the OO approach
Now that #107 is merged, we can work on the design of any class hierarchy & relationship needed for the rest of the code.

@Th3nn3ss just spent some time reading the code after your last changes in #107 , and from what I understand so far, I think (@mr-c feel free to chime-in and correct me if I said anything wrong, please :grimacing: )

- `miniwdl` uses its own classes when parsing WDL; so the `doc_tree` object that you created is an instance of [`WDL.Document` that contains a list of `WDL.Task`'s](https://github.com/chanzuckerberg/miniwdl/blob/0463f5a418974691562085f2ff1bfe4b4a411aa0/WDL/Tree.py#L1370)
- We can use `miniwdl` classes/objects
- Then we convert these into CWL objects, using the classes from `cwl_utils`

So I think we may need very few classes just for the code that handles parsing the WDL and creating the CWL file.

We suggested to start by sketching the classes and hierarchy & relationships that you have in mind. From your previous work, you should be able to imagine a few possible classes. Don't be afraid in suggesting some initial structure, even if you are not too sure, or if it has some missing links in the relationships (e.g. you are not sure how two classes relate to each other, but you think they must exist and be related some way).

## References

@Th3nn3ss good references for Python programming with (and without) OO:

- [Fluent Python](https://www.fluentpython.com/) (my local library has a copy, perhaps yours has a copy too?)
- Python 3.9 documentation for [classes](https://docs.python.org/3.9/tutorial/classes.html) (check out the other sections too)

Th3nn3ss:
Closes #109 

An Object Oriented implementation of the main_oop.py file. 
This implementation is solely meant to achieve the same results as the main_oop.py using the bowtie_1.wdl file.

- [x] First convert the  bowtie_1 wdl inputs to their equivalent cwl command input parameter object
- [x] Second convert the bowtie_1 wdl outputs to their equivalent cwl command output parameter object
- [x] Third, convert the bowtie_1 wdl runtime docker requirement to cwl Docker Requirement
- [x] Fourth, convert the bowtie_1 wdl commands to cwl Process Requirements
  - [x] convert wdl command part of type string
  - [x] handler proper indentation for translated cwl string
  - [x] convert wdl command part of type WDL.Expr.Placeholder
  - [x] convert wdl placeholder expression of type WDL.Expr.Apply
  - [x] convert wdl placeholder expression of type WDL.Expr.Get
  - [x] convert expression of type WDL.Expr.Apply with argument list of length 1
  - [x] convert expression of type WDL.Expr.Apply with argument list of length 2
  - [x] convert expression of type WDL.Expr.Appy with argument list of length 3
  - [x] convert expression of type WDL.Expr.Get with optional == True
- [ ] Write tests
- [ ] (others to come)


wdl2cwl/main_oop_2.py
```diff
@@ -0,0 +1,250 @@
...
+                    cwl_command_str = (
+            "$(inputs."
+            + placeholder_name
+            + " ? "
+            + '"{true_value}" : "{false_value}")'.format(
+                true_value=wdl_placeholder.options["true"],
+                false_value=wdl_placeholder.options["false"],
+            )
+            + ")"
+        )
```
kinow, line 117: @Th3nn3ss in today's meeting you mentioned this placeholder function was getting too long. It's really (really really really!) good that you are thinking about that while you write your code. It helps maintaining the code, and is extremely helpful to other developers when the code is simpler.

Maybe we could introduce f-strings here? I had a quick look and couldn't see a lot of f-strings in this file, so perhaps that's something that you would be interested in adopting too.

Here's how I think the code above could be refactored (needs testing):

```suggestion
                if "true" in wdl_placeholder.options:
                    true_value = wdl_placeholder.options["true"]
                    false_value = wdl_placeholder.options["false"]
                    cwl_command_str = f'$(inputs.{placeholder_name} ? "{true_value}" : "{false_value}")'
```

wdl2cwl/main_oop_2.py
```diff
@@ -0,0 +1,250 @@
...
+                    + arg_name
+                    + " ? "
+                    + '"{false_value}" : "{true_value}")'.format(
+                        true_value=wdl_placeholder.options["true"],
+                        false_value=wdl_placeholder.options["false"],
+                    )
+                )
+            if len(wdl_placeholder.expr.arguments) == 2:
+                arg_name, arg_value = wdl_placeholder.expr.arguments
+                arg_name, arg_value = arg_name.literal.value, arg_value.expr.name
```
kinow, line 134: I think we are creating `arg_name` and `arg_value` twice above :point_up:

kinow:
Great work so far @Th3nn3ss . Thanks for updating the description with the tasks, and let me know if you need any help.

wdl2cwl/main_oop_2.py
kinow: More about f-strings in Python:

- <https://realpython.com/python-f-strings/>
- <https://www.python.org/dev/peps/pep-0498/> (PEP-498 added in 3.6)

Th3nn3ss: That's very true. I should probably be more consistent with f-strings as it's more readable.

wdl2cwl/main_oop_2.py
Th3nn3ss: Am just reusing the variable names from line 133.

mr-c (changes requested):
Can the `inputs()` function from https://github.com/common-workflow-lab/wdl-cwl-translator/blob/e6cde4b9db59add7b2896c8a44e0e511f921f197/wdl2cwl/main.py#L45 be copied and used everywhere here?

Just use it as it. Don't worry about caching results of the regex evaluation; if that becomes a problem later then that is easily fixed.

Thanks.

wdl2cwl/old_main_oop.py
```diff
@@ -0,0 +1,470 @@
+"""Main entrypoint for WDL2CWL."""
```
mr-c, line 1: Good for your reference on your computer, but maybe shouldn't be part of this PR?

wdl2cwl/tests/oop_cwl_files/bowtie_1.cwl
```diff
@@ -60,25 +60,25 @@ requirements:
...
-            $(inputs.allowContain ? "--allow-contain" : ""))\
-            $(inputs.threads === null ? "" : "--threads " inputs.threads )\
-            $(inputs.samRG === null ? "" : "--sam-RG '" inputs.samRG )$(inputs.samRG ? "" : "'")\
-            $(inputs.indexFiles[0].replace( "(\.rev)?\.[0-9]\.ebwt$", "") )\
-            $(inputs.readsUpstream.map(function(el) {return el.path}).join(","))\
-                   | picard -Xmx$(inputs.picardXmx) SortSam \
-                   INPUT=/dev/stdin \
-                   OUTPUT=$(inputs.outputPath) \
-                   SORT_ORDER=coordinate \
-                   CREATE_INDEX=true
```
mr-c, line 81: Great, the indenting problem is fixed now :-)

mr-c:
@Th3nn3ss Let me know if you want me to review the mypy errors with you.

wdl2cwl/tests/oop_cwl_files/bowtie_1.cwl
Th3nn3ss: Yes. Apparently the string has to start from a new line `(\n)` else it can't be handled by textwrap.dedent. So I split the string from the first new line and dedent the right-split and concatenate it to the left-split.

Th3nn3ss:
> Can the `inputs()` function from
> 
> https://github.com/common-workflow-lab/wdl-cwl-translator/blob/e6cde4b9db59add7b2896c8a44e0e511f921f197/wdl2cwl/main.py#L45
> 
> be copied and used everywhere here?
> Just use it as it. Don't worry about caching results of the regex evaluation; if that becomes a problem later then that is easily fixed.
> 
> Thanks.

This implementation is actually not using any data structure to store the inputs and their values. All information used are returned from the the individual WDL objects.

mr-c:
> > Can the `inputs()` function from
> > https://github.com/common-workflow-lab/wdl-cwl-translator/blob/e6cde4b9db59add7b2896c8a44e0e511f921f197/wdl2cwl/main.py#L45
> > 
> > be copied and used everywhere here?
> > Just use it as it. Don't worry about caching results of the regex evaluation; if that becomes a problem later then that is easily fixed.
> > Thanks.
> 
> This implementation is actually not using any data structure to store the inputs and their values. All information used are returned from the the individual WDL objects.

`inputs()` is not needed for `bowtie_1.wdl` but will be need for any WDL file that has an input name that is not a valid javascript identifier

`inputs["identifier"]` (always safe, a bit verbose) versus `inputs.identifier` (nicer looking, not allowed for some identifiers)

So it doesn't have to be part of this PR. If you don't want to do it now, then please make an issue to remind yourself

Th3nn3ss:
> > > Can the `inputs()` function from
> > > https://github.com/common-workflow-lab/wdl-cwl-translator/blob/e6cde4b9db59add7b2896c8a44e0e511f921f197/wdl2cwl/main.py#L45
> > > 
> > > be copied and used everywhere here?
> > > Just use it as it. Don't worry about caching results of the regex evaluation; if that becomes a problem later then that is easily fixed.
> > > Thanks.
> > 
> > 
> > This implementation is actually not using any data structure to store the inputs and their values. All information used are returned from the the individual WDL objects.
> 
> `inputs()` is not needed for `bowtie_1.wdl` but will be need for any WDL file that has an input name that is not a valid javascript identifier
> 
> `inputs["identifier"]` (always safe, a bit verbose) versus `inputs.identifier` (nicer looking, not allowed for some identifiers)
> 
> So it doesn't have to be part of this PR. If you don't want to do it now, then please make an issue to remind yourself

Oh..ok. I haven't seen that yet that's why I don't understand what you mean. I will make an issue for it.

Th3nn3ss:
> @Th3nn3ss Let me know if you want me to review the mypy errors with you. @mr-c yes please
@mr-c yes please

wdl2cwl/main_oop.py
```diff
@@ -15,571 +14,331 @@
...
+                        + f'"{arg_name}"'
+                        + " inputs."
+                        + arg_value
+                        + " )"
+                    )
+                else:
+                    cwl_command_str = f"{arg_name} $(inputs.{arg_value})"
+            elif function_name == "sub":
+                wdl_apply_object_arg, arg_string, arg_substitute = expr_arguments
+                if not hasattr(wdl_apply_object_arg, "arguments"):
```
mr-c, line 211: What type of `WDL` object from miniwdl has an attribute named `arguments`? Can we test for that type instead?

Th3nn3ss: That will be WDL.Expr.Apply

mr-c (approved):
Good enough to move forward; thanks!

Issue #114, opened 2021-12-21
Simplify translate_wdl_placeholder


kinow:
@Th3nn3ss mentioned on Element the function `translate_wdl_placeholder`, which could be simplified. Below some suggestions for later.

```py
    def translate_wdl_placeholder(self, wdl_placeholder: WDL.Expr.Placeholder) -> str:
        """Translate WDL Expr Placeholder to a valid CWL command string."""
        # 1. you can use a defaultdict to simplify it?
        options = defaultdict(lambda: None, wdl_placeholder.options)
        true_value = options["true"]
        false_value = options["false"]
        separator = options["sep"]
        # 2. this can go here, or in the class constructor; adding new types, means updating a dictionary,
        # instead of writing another `elif blabla: call_some_other_method` which gets longer and tedious
        placeholder_handlers = dict({
            WDL.Expr.Get.__name__: "_handle_placeholder_get",
            WDL.Expr.Apply.__name__: "_handle_placeholder_apply"
        })
        try:
            # 3. getattr gets an attribute of an object, in this case, `self` is the object, and the
            # attribute is the handler for the WDL Expr type; we also pass the options values.
            value : str = getattr(self, placeholder_handlers[wdl_placeholder.expr.__class__.__name__])(
                wdl_placeholder.expr,
                true_value=true_value,
                false_value=false_value,
                separator=separator)
            return value
        except KeyError:
            # 4. an invalid value? not sure how to handle it, probably an error or log?
            raise ValueError('Invalid placeholder!')

    def _handle_placeholder_get(self, nested_expr, *, true_value, false_value, separator) -> str:
        if nested_expr is None or not isinstance(nested_expr, WDL.Expr.Ident):
            raise ValueError(f"Unsupported type: {type(nested_expr)}")
        ...
        return ''

    def _handle_placeholder_apply(self, wdl_placeholder_expr, *, true_value, false_value, separator) -> str:
        return ''
```

See <https://realpython.com/python-is-identity-vs-equality/> for why we need to use `is None` or `is not None`.

See <https://docs.python.org/3/library/collections.html#collections.defaultdict> for more about the `defaultdict` from the Python stdlib.

PR #116, opened 2021-12-21
start bcftools_stats.wdl adoptation with main_oop

### LINKED ISSUE #113: add inputs( ) function to OOP implementation
Use the `inputs()` function from
https://github.com/common-workflow-lab/wdl-cwl-translator/blob/e6cde4b9db59add7b2896c8a44e0e511f921f197/wdl2cwl/main.py#L45

This will be need for any WDL file that has an input name that is not a valid javascript identifier.

`inputs["identifier"]` (always safe, a bit verbose) versus `inputs.identifier` (nicer looking, not allowed for some identifiers)

Th3nn3ss:
Closes #113 #114 
This is an attempt to run bcftools_stats.wdl through the main_oop.py.

The file is hard coded to the main_oop.py file and some of the cwl.CommandLineTool arguments are left in default and processed one after the other. The order of tasks to be carried out are:

- [x] Handle bcftools_stats inputs and convert to cwl inputs
  - [x] Handle inputs with optional false
  - [x] Handle inputs with WDL.Expr.Apply expressions
  - [x] Handle inputs with WDL.Type.Array and item_type of WDL.Type.String
- [x] Handle bcftools_stats outputs and convert to cwl outputs
- [x] Handle bcftools_stats command and convert to cwl Process Requirement
    - [x] Handle command with expression inputs since cwl inputs don't default to expressions 
- [x] Handle bcftools_stats runtime requirements
    - [x] Handle runtime cpu requirements with WDL.Expr.Apply expressions
    - [x] Handle runtime memory requirements
- [x] Run tests


wdl2cwl/main_oop.py
```diff
@@ -250,46 +252,61 @@ def get_cwl_inputs(
         """Convert WDL inputs into CWL inputs and return a list of CWL Command Input Paramenters."""
         inputs: List[cwl.CommandInputParameter] = []
 
-        if not wdl_inputs:
-            return inputs
-        for wdl_input in wdl_inputs:
```
mr-c, line 255: Can you keep this simplification? It is nice to not have to indent an extra level

wdl2cwl/main_oop.py
```diff
@@ -337,7 +354,7 @@ def main() -> None:
     #     with open(args.output, "w") as result:
     # result.write(str(Converter.load_wdl_tree("wdl2cwl/tests/wdl_files/bcftools_stats.wdl"))) #missing args.workflow)
 
-    Converter.load_wdl_tree("wdl2cwl/tests/wdl_files/bowtie_1.wdl")
+    Converter.load_wdl_tree("wdl2cwl/tests/wdl_files/bcftools_stats.wdl")
```
kinow, line 357: @Th3nn3ss maybe we can leave both calls above (it's fine if you add it back before we review&merge this PR). Once we get these 2 workflows being translated correctly, then we can start adding more workflows making sure we didn't break others.

Later this will be replaced when we switch the old translator to this one, and use it in the unit tests :+1:

Th3nn3ss: both file path will be added to test_miniwdl.py at the end of this PR. Pytest will be conducted on both files.

kinow: Excellent @Th3nn3ss ! Thanks

wdl2cwl/tests/oop_cwl_files/bcftools_stats.cwl
```diff
@@ -0,0 +1,170 @@
...
+            $(inputs.targetsFile === null ? "" : "--targets-file " inputs.targetsFile) \
+            $(inputs.userTsTv === null ? "" : "--user-tstv " inputs.userTsTv) \
+            --threads $(inputs.threads) \
+            $(inputs.verbose ? "--verbose" : "") \
+            $(inputs.inputVcf)$(inputs.compareVcf) > $(inputs.outputPath === null ? inputs.inputVcf.basename + '.stats' : inputs.outputPath)
+  - class: InlineJavascriptRequirement
+  - class: NetworkAccess
+    networkAccess: true
+  - class: ResourceRequirement
+    ramMin: inputs.threads + 1
```
mr-c, line 149: double with then next line? Also, not a float nor a CWL expression

Th3nn3ss: Sorry about that.

wdl2cwl/main_oop.py
```diff
@@ -51,30 +51,28 @@ def load_wdl_task(self, obj: WDL.Tree.Task) -> str:
...
-        # if not isinstance(runtime_docker, WDL.Expr.Get):
-        #     raise Exception(
-        #         f"Unsupport docker runtime type: {type(runtime_docker)}: {runtime_docker}"
-        #     )
-        # docker_requirement = self.get_cwl_docker_requirements(runtime_docker)
-        # cwl_command_str = self.get_cwl_command_requirements(obj.command.parts)
+        runtime_docker = obj.runtime["docker"]
+        if not isinstance(runtime_docker, WDL.Expr.Get):
+            raise Exception(
+                f"Unsupport docker runtime type: {type(runtime_docker)}: {runtime_docker}"
```
kinow, line 57: s/Unsupport/Unsupported (I think)

wdl2cwl/tests/oop_cwl_files/bcftools_stats.cwl
```diff
@@ -0,0 +1,170 @@
...
+            --threads $(inputs.threads) \
+            $(inputs.verbose ? "--verbose" : "") \
+            $(inputs.inputVcf)$(inputs.compareVcf) > $(inputs.outputPath === null ? inputs.inputVcf.basename + '.stats' : inputs.outputPath)
+  - class: InlineJavascriptRequirement
+  - class: NetworkAccess
+    networkAccess: true
+  - class: ResourceRequirement
+    coresMin: $(inputs.threads + 1)
+  - class: ResourceRequirement
+    ramMin: |-
```
mr-c, line 151: There should be a single `ResourceRequirement`, with both `coresMin` and `ramMin`

kinow (commented):
@Th3nn3ss `mypy` passes for me with the following diff:

```diff
diff --git a/wdl2cwl/main_oop.py b/wdl2cwl/main_oop.py
index 7551046..2498c14 100644
--- a/wdl2cwl/main_oop.py
+++ b/wdl2cwl/main_oop.py
@@ -67,7 +67,7 @@ class Converter:
         requirements.append(cwl.NetworkAccess(networkAccess=True))
         cpu_requirement = self.get_cpu_requirement(obj.runtime["cpu"])
         requirements.append(cpu_requirement)
-        if "memory" in obj.runtime:
+        if "memory" in obj.runtime and isinstance(obj.runtime["memory"], WDL.Expr.Placeholder):
             memory_requirement = self.get_memory_requirement(obj.runtime["memory"])
             requirements.append(memory_requirement)
 
@@ -93,14 +93,15 @@ class Converter:
         return result_stream.getvalue()
 
     def get_memory_requirement(
-        self, memory_runtime: WDL.Expr.Base
+        self, memory_runtime: WDL.Expr.Placeholder
     ) -> cwl.ResourceRequirement:
         """Translate WDL Runtime Memory requirement to CWL Resource Requirement."""
         ram_min = ""
         if isinstance(memory_runtime.expr, WDL.Expr.Ident):
             expr_referee = memory_runtime.expr.referee
-            expr_referee_name = expr_referee.name
-            ram_min = self.get_ram_min_js(expr_referee_name, None)
+            if expr_referee:
+                expr_referee_name = str(expr_referee.name)
+                ram_min = self.get_ram_min_js(expr_referee_name, '')
         return cwl.ResourceRequirement(ramMin=ram_min)
 
     def get_ram_min_js(self, ram_min_ref_name: str, unit: str) -> str:
@@ -147,7 +148,7 @@ class Converter:
             if ref_function == "_add":
                 first_arg, second_arg = ref_arguments
                 second_arg_value = self.get_wdl_literal(second_arg.literal)
-                first_arg_expr_name = first_arg.expr.name
+                first_arg_expr_name = cast(WDL.Expr.Ident, cast(WDL.Expr.Get, first_arg).expr).name
                 cores_min = f"$(inputs.{first_arg_expr_name} + {second_arg_value})"
 
         else:
@@ -411,7 +412,10 @@ class Converter:
             if isinstance(wdl_output.type, WDL.Type.File):
                 type_of = "File"
             # check for output with referee and referee expr of type WDL.Expr.Apply
-            expr_ident = wdl_output.expr.expr
+            if not wdl_output.expr:
+                raise ValueError('Missing expression')
+            get_expr = cast(WDL.Expr.Get, wdl_output.expr)
+            expr_ident = cast(WDL.Expr.Ident, get_expr.expr)
             expr_ident_name = expr_ident.name
             if expr_ident.referee and isinstance(
                 expr_ident.referee.expr, WDL.Expr.Apply
@@ -437,7 +441,7 @@ class Converter:
                     + f" ({true_tenary}) : inputs.{reference_expr.name})"
                 )
             else:
-                glob_expr = f"$(inputs.{expr_ident_name})"  # type: ignore[union-attr]
+                glob_expr = f"$(inputs.{expr_ident_name})"
 
             outputs.append(
                 cwl.CommandOutputParameter(
```

I don't know `miniwdl` well enough to say it's 100% correct, but here's how I solved the issues:

- ran `mypy wdl2cwl`
- set breakpoints in the lines where there were errors
- executed the code with the new workflow you chose
- hovered the variable, then clicked on Go to Source (in PyCharm)
- Looked at the pydocs, and class hierarchy
- Picked a type

In the cases where `mypy` complains about `Base` not having `expr`, it's really tricky. In `miniwdl` we have a list of `Base` objects (i.e. they are instances of subclasses of `Base`, since [`Base` is abstract](https://docs.python.org/3/library/abc.html)).

The only two subclasses where I found the `.expr` attribute were `Get` and `Placeholder`. I used the types that the debugger found, but I think there's no hard-constraint on the type of that list, and we could have a `Placeholder` somewhere by accident. That would not break our code, so I think we should be good to go with this diff?

-Bruno

Th3nn3ss:
> @Th3nn3ss `mypy` passes for me with the following diff:
> 
> ```diff
> diff --git a/wdl2cwl/main_oop.py b/wdl2cwl/main_oop.py
> index 7551046..2498c14 100644
> --- a/wdl2cwl/main_oop.py
> +++ b/wdl2cwl/main_oop.py
> @@ -67,7 +67,7 @@ class Converter:
>          requirements.append(cwl.NetworkAccess(networkAccess=True))
>          cpu_requirement = self.get_cpu_requirement(obj.runtime["cpu"])
>          requirements.append(cpu_requirement)
> -        if "memory" in obj.runtime:
> +        if "memory" in obj.runtime and isinstance(obj.runtime["memory"], WDL.Expr.Placeholder):
>              memory_requirement = self.get_memory_requirement(obj.runtime["memory"])
>              requirements.append(memory_requirement)
>  
> @@ -93,14 +93,15 @@ class Converter:
>          return result_stream.getvalue()
>  
>      def get_memory_requirement(
> -        self, memory_runtime: WDL.Expr.Base
> +        self, memory_runtime: WDL.Expr.Placeholder
>      ) -> cwl.ResourceRequirement:
>          """Translate WDL Runtime Memory requirement to CWL Resource Requirement."""
>          ram_min = ""
>          if isinstance(memory_runtime.expr, WDL.Expr.Ident):
>              expr_referee = memory_runtime.expr.referee
> -            expr_referee_name = expr_referee.name
> -            ram_min = self.get_ram_min_js(expr_referee_name, None)
> +            if expr_referee:
> +                expr_referee_name = str(expr_referee.name)
> +                ram_min = self.get_ram_min_js(expr_referee_name, '')
>          return cwl.ResourceRequirement(ramMin=ram_min)
>  
>      def get_ram_min_js(self, ram_min_ref_name: str, unit: str) -> str:
> @@ -147,7 +148,7 @@ class Converter:
>              if ref_function == "_add":
>                  first_arg, second_arg = ref_arguments
>                  second_arg_value = self.get_wdl_literal(second_arg.literal)
> -                first_arg_expr_name = first_arg.expr.name
> +                first_arg_expr_name = cast(WDL.Expr.Ident, cast(WDL.Expr.Get, first_arg).expr).name
>                  cores_min = f"$(inputs.{first_arg_expr_name} + {second_arg_value})"
>  
>          else:
> @@ -411,7 +412,10 @@ class Converter:
>              if isinstance(wdl_output.type, WDL.Type.File):
>                  type_of = "File"
>              # check for output with referee and referee expr of type WDL.Expr.Apply
> -            expr_ident = wdl_output.expr.expr
> +            if not wdl_output.expr:
> +                raise ValueError('Missing expression')
> +            get_expr = cast(WDL.Expr.Get, wdl_output.expr)
> +            expr_ident = cast(WDL.Expr.Ident, get_expr.expr)
>              expr_ident_name = expr_ident.name
>              if expr_ident.referee and isinstance(
>                  expr_ident.referee.expr, WDL.Expr.Apply
> @@ -437,7 +441,7 @@ class Converter:
>                      + f" ({true_tenary}) : inputs.{reference_expr.name})"
>                  )
>              else:
> -                glob_expr = f"$(inputs.{expr_ident_name})"  # type: ignore[union-attr]
> +                glob_expr = f"$(inputs.{expr_ident_name})"
>  
>              outputs.append(
>                  cwl.CommandOutputParameter(
> ```
> 
> I don't know `miniwdl` well enough to say it's 100% correct, but here's how I solved the issues:
> 
> * ran `mypy wdl2cwl`
> * set breakpoints in the lines where there were errors
> * executed the code with the new workflow you chose
> * hovered the variable, then clicked on Go to Source (in PyCharm)
> * Looked at the pydocs, and class hierarchy
> * Picked a type
> 
> In the cases where `mypy` complains about `Base` not having `expr`, it's really tricky. In `miniwdl` we have a list of `Base` objects (i.e. they are instances of subclasses of `Base`, since [`Base` is abstract](https://docs.python.org/3/library/abc.html)).
> 
> The only two subclasses where I found the `.expr` attribute were `Get` and `Placeholder`. I used the types that the debugger found, but I think there's no hard-constraint on the type of that list, and we could have a `Placeholder` somewhere by accident. That would not break our code, so I think we should be good to go with this diff?
> 
> -Bruno

Awesome

Th3nn3ss:
> @Th3nn3ss `mypy` passes for me with the following diff:
> 
> ```diff
> diff --git a/wdl2cwl/main_oop.py b/wdl2cwl/main_oop.py
> index 7551046..2498c14 100644
> --- a/wdl2cwl/main_oop.py
> +++ b/wdl2cwl/main_oop.py
> @@ -67,7 +67,7 @@ class Converter:
>          requirements.append(cwl.NetworkAccess(networkAccess=True))
>          cpu_requirement = self.get_cpu_requirement(obj.runtime["cpu"])
>          requirements.append(cpu_requirement)
> -        if "memory" in obj.runtime:
> +        if "memory" in obj.runtime and isinstance(obj.runtime["memory"], WDL.Expr.Placeholder):
>              memory_requirement = self.get_memory_requirement(obj.runtime["memory"])
>              requirements.append(memory_requirement)
>  
> @@ -93,14 +93,15 @@ class Converter:
>          return result_stream.getvalue()
>  
>      def get_memory_requirement(
> -        self, memory_runtime: WDL.Expr.Base
> +        self, memory_runtime: WDL.Expr.Placeholder
>      ) -> cwl.ResourceRequirement:
>          """Translate WDL Runtime Memory requirement to CWL Resource Requirement."""
>          ram_min = ""
>          if isinstance(memory_runtime.expr, WDL.Expr.Ident):
>              expr_referee = memory_runtime.expr.referee
> -            expr_referee_name = expr_referee.name
> -            ram_min = self.get_ram_min_js(expr_referee_name, None)
> +            if expr_referee:
> +                expr_referee_name = str(expr_referee.name)
> +                ram_min = self.get_ram_min_js(expr_referee_name, '')
>          return cwl.ResourceRequirement(ramMin=ram_min)
>  
>      def get_ram_min_js(self, ram_min_ref_name: str, unit: str) -> str:
> @@ -147,7 +148,7 @@ class Converter:
>              if ref_function == "_add":
>                  first_arg, second_arg = ref_arguments
>                  second_arg_value = self.get_wdl_literal(second_arg.literal)
> -                first_arg_expr_name = first_arg.expr.name
> +                first_arg_expr_name = cast(WDL.Expr.Ident, cast(WDL.Expr.Get, first_arg).expr).name
>                  cores_min = f"$(inputs.{first_arg_expr_name} + {second_arg_value})"
>  
>          else:
> @@ -411,7 +412,10 @@ class Converter:
>              if isinstance(wdl_output.type, WDL.Type.File):
>                  type_of = "File"
>              # check for output with referee and referee expr of type WDL.Expr.Apply
> -            expr_ident = wdl_output.expr.expr
> +            if not wdl_output.expr:
> +                raise ValueError('Missing expression')
> +            get_expr = cast(WDL.Expr.Get, wdl_output.expr)
> +            expr_ident = cast(WDL.Expr.Ident, get_expr.expr)
>              expr_ident_name = expr_ident.name
>              if expr_ident.referee and isinstance(
>                  expr_ident.referee.expr, WDL.Expr.Apply
> @@ -437,7 +441,7 @@ class Converter:
>                      + f" ({true_tenary}) : inputs.{reference_expr.name})"
>                  )
>              else:
> -                glob_expr = f"$(inputs.{expr_ident_name})"  # type: ignore[union-attr]
> +                glob_expr = f"$(inputs.{expr_ident_name})"
>  
>              outputs.append(
>                  cwl.CommandOutputParameter(
> ```
> 
> I don't know `miniwdl` well enough to say it's 100% correct, but here's how I solved the issues:
> 
> * ran `mypy wdl2cwl`
> * set breakpoints in the lines where there were errors
> * executed the code with the new workflow you chose
> * hovered the variable, then clicked on Go to Source (in PyCharm)
> * Looked at the pydocs, and class hierarchy
> * Picked a type
> 
> In the cases where `mypy` complains about `Base` not having `expr`, it's really tricky. In `miniwdl` we have a list of `Base` objects (i.e. they are instances of subclasses of `Base`, since [`Base` is abstract](https://docs.python.org/3/library/abc.html)).
> 
> The only two subclasses where I found the `.expr` attribute were `Get` and `Placeholder`. I used the types that the debugger found, but I think there's no hard-constraint on the type of that list, and we could have a `Placeholder` somewhere by accident. That would not break our code, so I think we should be good to go with this diff?
> 
> -Bruno

This is amazing. This is definitely handy. I will have to explore it in due time.

Th3nn3ss:
> @Th3nn3ss `mypy` passes for me with the following diff:
> 
> ```diff
> diff --git a/wdl2cwl/main_oop.py b/wdl2cwl/main_oop.py
> index 7551046..2498c14 100644
> --- a/wdl2cwl/main_oop.py
> +++ b/wdl2cwl/main_oop.py
> @@ -67,7 +67,7 @@ class Converter:
>          requirements.append(cwl.NetworkAccess(networkAccess=True))
>          cpu_requirement = self.get_cpu_requirement(obj.runtime["cpu"])
>          requirements.append(cpu_requirement)
> -        if "memory" in obj.runtime:
> +        if "memory" in obj.runtime and isinstance(obj.runtime["memory"], WDL.Expr.Placeholder):
>              memory_requirement = self.get_memory_requirement(obj.runtime["memory"])
>              requirements.append(memory_requirement)
>  
> @@ -93,14 +93,15 @@ class Converter:
>          return result_stream.getvalue()
>  
>      def get_memory_requirement(
> -        self, memory_runtime: WDL.Expr.Base
> +        self, memory_runtime: WDL.Expr.Placeholder
>      ) -> cwl.ResourceRequirement:
>          """Translate WDL Runtime Memory requirement to CWL Resource Requirement."""
>          ram_min = ""
>          if isinstance(memory_runtime.expr, WDL.Expr.Ident):
>              expr_referee = memory_runtime.expr.referee
> -            expr_referee_name = expr_referee.name
> -            ram_min = self.get_ram_min_js(expr_referee_name, None)
> +            if expr_referee:
> +                expr_referee_name = str(expr_referee.name)
> +                ram_min = self.get_ram_min_js(expr_referee_name, '')
>          return cwl.ResourceRequirement(ramMin=ram_min)
>  
>      def get_ram_min_js(self, ram_min_ref_name: str, unit: str) -> str:
> @@ -147,7 +148,7 @@ class Converter:
>              if ref_function == "_add":
>                  first_arg, second_arg = ref_arguments
>                  second_arg_value = self.get_wdl_literal(second_arg.literal)
> -                first_arg_expr_name = first_arg.expr.name
> +                first_arg_expr_name = cast(WDL.Expr.Ident, cast(WDL.Expr.Get, first_arg).expr).name
>                  cores_min = f"$(inputs.{first_arg_expr_name} + {second_arg_value})"
>  
>          else:
> @@ -411,7 +412,10 @@ class Converter:
>              if isinstance(wdl_output.type, WDL.Type.File):
>                  type_of = "File"
>              # check for output with referee and referee expr of type WDL.Expr.Apply
> -            expr_ident = wdl_output.expr.expr
> +            if not wdl_output.expr:
> +                raise ValueError('Missing expression')
> +            get_expr = cast(WDL.Expr.Get, wdl_output.expr)
> +            expr_ident = cast(WDL.Expr.Ident, get_expr.expr)
>              expr_ident_name = expr_ident.name
>              if expr_ident.referee and isinstance(
>                  expr_ident.referee.expr, WDL.Expr.Apply
> @@ -437,7 +441,7 @@ class Converter:
>                      + f" ({true_tenary}) : inputs.{reference_expr.name})"
>                  )
>              else:
> -                glob_expr = f"$(inputs.{expr_ident_name})"  # type: ignore[union-attr]
> +                glob_expr = f"$(inputs.{expr_ident_name})"
>  
>              outputs.append(
>                  cwl.CommandOutputParameter(
> ```
> 
> I don't know `miniwdl` well enough to say it's 100% correct, but here's how I solved the issues:
> 
> * ran `mypy wdl2cwl`
> * set breakpoints in the lines where there were errors
> * executed the code with the new workflow you chose
> * hovered the variable, then clicked on Go to Source (in PyCharm)
> * Looked at the pydocs, and class hierarchy
> * Picked a type
> 
> In the cases where `mypy` complains about `Base` not having `expr`, it's really tricky. In `miniwdl` we have a list of `Base` objects (i.e. they are instances of subclasses of `Base`, since [`Base` is abstract](https://docs.python.org/3/library/abc.html)).
> 
> The only two subclasses where I found the `.expr` attribute were `Get` and `Placeholder`. I used the types that the debugger found, but I think there's no hard-constraint on the type of that list, and we could have a `Placeholder` somewhere by accident. That would not break our code, so I think we should be good to go with this diff?
> 
> -Bruno

`if "memory" in obj.runtime and isinstance(obj.runtime["memory"], WDL.Expr.Placeholder):` This actually doesn't allow runtime memory to be translated because obj.runtime["memory"] is of instance WDL.Expr.Get

wdl2cwl/tests/oop_cwl_files/bcftools_stats.cwl
Th3nn3ss: This isn't consistent with other cwl files. How do I achieve this?

kinow:
>if "memory" in obj.runtime and isinstance(obj.runtime["memory"], WDL.Expr.Placeholder): This actually doesn't allow runtime memory to be translated because obj.runtime["memory"] is of instance WDL.Expr.Get

Ah, my bad. I think I got confused while reading the code for Placeholder and Get. Feel free to modify the diff as you see fit @Th3nn3ss :+1:

Th3nn3ss:
I don't have tests to cover the newly created exceptions. Should I be bothered about passing codecov/patch? @kinow

wdl2cwl/tests/oop_cwl_files/bcftools_stats.cwl
kinow: Each `- class: ResourceRequirement` is an YAML node/object. `requirements` is a collection/list (see how each of its children in this YAML start with a dash `-`?).

So I think here we have two entries in this list that define `- class: ResourceRequirement`, which in CWL would mean you are defining that twice. Not sure if it works, probably depends on implementation?

Anyway, there's an example here in the user guide with both `coreMin` and `ramMin` in the same `ResourceRequirement` object: https://www.commonwl.org/user_guide/20-software-requirements/index.html

If you'd like to see how the YAML document is parsed, @Th3nn3ss , you can try this. Go to this URL <https://codebeautify.org/yaml-parser-online>. Then paste this simplified version of this YAML.

```yaml
class: CommandLineTool
# simplify
requirements:
  - class: ResourceRequirement
    coresMin: $(inputs.threads + 1)
  - class: ResourceRequirement
    ramMin: |-
        ${
        var unit = inputs.memory.match(/[a-zA-Z]+/g).join("");
        }
```

It should show the two `ResourceRequirement`'s under `requirements (we want one). You can play with that online editor and try other things to see what the YAML output will be.

image

kinow (approved):
This code is looking great @Th3nn3ss ! CI is passing, I found nothing that looked wrong in the code. Leaving my +1 here. Thank you!

kinow:
> I don't have tests to cover the newly created exceptions. Should I be bothered about passing codecov/patch? @kinow

I think the code might change more as we add more workflows to the test. So I would worry about coverage when we are getting ready to the final version of the new parser. But better to confirm with @mr-c :+1:

mr-c (changes requested):
Please fix the errors indicated by `cwltool wdl2cwl/tests/oop_cwl_files/bcftools_stats.cwl wdl2cwl/tests/inputs/bcftools_stats.yaml`

```
79 (function(){return ((inputs.afBins === null ? "" : "--af-bins " inputs.afBins));})()
stdout was: ''
stderr was: 'evalmachine.<anonymous>:79
(function(){return ((inputs.afBins === null ? "" : "--af-bins " inputs.afBins));})()
                                                                ^^^^^^

SyntaxError: Unexpected identifier
```

wdl2cwl/tests/oop_cwl_files/bcftools_stats.cwl
```diff
@@ -0,0 +1,169 @@
...
+  - class: InitialWorkDirRequirement
+    listing:
+      - entryname: example.sh
+        entry: |4
+
+            set -e
+            mkdir -p \$(dirname  $(inputs.outputPath === null ? inputs.inputVcf.basename + '.stats' : inputs.outputPath))
+            bcftools stats \
+            $(inputs.afBins === null ? "" : "--af-bins " inputs.afBins) \
+            $(inputs.afTag === null ? "" : "--af-tag " inputs.afTag) \
```
mr-c, line 125:
```suggestion
            $(inputs.afBins === null ? "" : "--af-bins " + inputs.afBins) \
            $(inputs.afTag === null ? "" : "--af-tag " + inputs.afTag) \
```

mr-c:
Also try `meld wdl2cwl/tests/oop_cwl_files/bcftools_stats.cwl wdl2cwl/tests/cwl_files/bcftools_stats.cwl` to see the differences between the old and new

wdl2cwl/tests/test_miniwdl.py
```diff
@@ -15,7 +15,7 @@ def get_file(path: str) -> str:
 @pytest.mark.parametrize(
     "wdl_path,cwl_path",
     [
-        # ("wdl_files/bowtie_1.wdl", "oop_cwl_files/bowtie_1.cwl"),
+        ("wdl_files/bowtie_1.wdl", "oop_cwl_files/bowtie_1.cwl"),
```
mr-c, line 18: ? I thought this PR is about bcftools stats?

wdl2cwl/tests/oop_cwl_files/bcftools_stats.cwl
```diff
@@ -0,0 +1,167 @@
+class: CommandLineTool
+id: Stats
+inputs:
+  - id: inputVcf
+    type: File
+  - id: inputVcfIndex
+    type: File
+  - id: outputPath
+    default: basename(inputVcf) + .stats
```
mr-c, line 9: This is not a valid CWL default, sorry. We can't use expressions here. This is an error in the original `bcftools_stats`

wdl2cwl/tests/oop_cwl_files/bcftools_stats.cwl
```diff
@@ -0,0 +1,167 @@
...
+    default: 256M
+    type: string
+  - id: dockerImage
+    default: quay.io/biocontainers/bcftools:1.10.2--h4f4756c_2
+    type: string
+outputs:
+  - id: stats
+    type: File
+    outputBinding:
+        glob: $(inputs.outputPath)
```
mr-c, line 110:
```suggestion
        glob: $(inputs.outputPath === null ? inputs.inputVcf.basename + ".stats" : inputs.outputPath)
```

mr-c (changes requested):

wdl2cwl/tests/test_miniwdl.py
Th3nn3ss: This was to pass the bowtie_1 that was already in the test. Both bowtie_1 and bcftools_stats had the sep function which is the equivalent of the join function, fixing the function for bcftools_stats caused error for bowtie_1.  I later discovered that although the join separator was used for arrays it mattered the content of the items of the array.

wdl2cwl/tests/oop_cwl_files/bcftools_stats.cwl
Th3nn3ss: Ok. I will revert to the former solution you suggested.

mr-c (approved):

wdl2cwl/tests/oop_cwl_files/bcftools_stats.cwl
```diff
@@ -0,0 +1,173 @@
...
+    type: string
+  - id: dockerImage
+    default: quay.io/biocontainers/bcftools:1.10.2--h4f4756c_2
+    type: string
+outputs:
+  - id: stats
+    type: File
+    outputBinding:
+        glob: "$(inputs.outputPath == null ? inputs.inputVcf.basename + '.stats' :\
+            \ inputs.outputPath)"
```
mr-c, line 112:
```suggestion
            inputs.outputPath)"
```

wdl2cwl/tests/oop_cwl_files/bcftools_stats.cwl
```diff
@@ -0,0 +1,173 @@
...
+    listing:
+      - entryname: example.sh
+        entry: |4
+
+            set -e
+            mkdir -p \$(dirname $(inputs.outputPath == null ? inputs.inputVcf.basename + '.stats' : inputs.outputPath))
+            mkdir fastaRef_dir
+            ln -s $(inputs.fastaRef == null ? "" : inputs.fastaRef.path) fastaRef_dir/\$(basename $(inputs.fastaRef == null ? "" : inputs.fastaRef.path))
+            ln -s $(inputs.fastaRefIndex == null ? "" : inputs.fastaRefIndex.path) fastaRef_dir/\$(basename $(inputs.fastaRefIndex == null ? "" : inputs.fastaRefIndex.path))
+            bcftools stats \
```
mr-c, line 126:
```suggestion
            ln -s $(inputs.fastaRef === null ? "" : inputs.fastaRef.path) fastaRef_dir/\$(basename $(inputs.fastaRef === null ? "" : inputs.fastaRef.path))
            ln -s $(inputs.fastaRefIndex === null ? "" : inputs.fastaRefIndex.path) fastaRef_dir/\$(basename $(inputs.fastaRefIndex === null ? "" : inputs.fastaRefIndex.path))
            bcftools stats \
```

wdl2cwl/main_oop.py
```diff
@@ -89,15 +111,213 @@ def load_wdl_task(self, obj: WDL.Tree.Task) -> str:
...
+            left_operand, right_operand = arguments
+            right_operand = self.get_wdl_literal(right_operand.literal)  # type: ignore
+            left_operand_value = self.get_expr(left_operand)
+            if getattr(left_operand, "function_name", None) == "basename":
+                treat_as_optional = True
+                referer = wdl_apply_expr.parent.name  # type: ignore
+            return (
+                f"{left_operand_value} + {right_operand}"
+                if not treat_as_optional
+                else f"{self.get_input(referer)} == null ? {left_operand_value} + '{right_operand}' : {self.get_input(referer)}"
```
mr-c, line 216:
```suggestion
                else f"{self.get_input(referer)} === null ? {left_operand_value} + '{right_operand}' : {self.get_input(referer)}"
```

wdl2cwl/main_oop.py
```diff
@@ -89,15 +111,213 @@ def load_wdl_task(self, obj: WDL.Tree.Task) -> str:
...
+                and not isinstance(wdl_get_expr.type, WDL.Type.File)
+            ):
+                return self.get_expr(referee.expr)
+            if optional and isinstance(wdl_get_expr.type, WDL.Type.File):
+                # To prevent null showing on the terminal for inputs of type File
+                just_id_name = self.get_expr_name(wdl_get_expr.expr)
+                with_file_check = self.get_expr_name_with_is_file_check(
+                    wdl_get_expr.expr
+                )
+                ident_name = f'{just_id_name} == null ? "" : {with_file_check}'
```
mr-c, line 294:
```suggestion
                ident_name = f'{just_id_name} === null ? "" : {with_file_check}'
```

wdl2cwl/main_oop.py
```diff
@@ -132,111 +352,90 @@ def translate_wdl_placeholder(self, wdl_placeholder: WDL.Expr.Placeholder) -> st
...
+                if isinstance(wdl_placeholder.expr, WDL.Expr.Get):
+                    is_optional = wdl_placeholder.expr.type.optional
+                elif isinstance(wdl_placeholder.expr, WDL.Expr.Apply):
+                    is_optional = wdl_placeholder.expr.arguments[0].type.optional
+                if not is_optional:
+                    cwl_command_str = (
+                        f'$({placeholder_expr} ? "{true_value}" : "{false_value}")'
+                    )
+                else:
+                    cwl_command_str = f'$({placeholder_expr} == null ? "{false_value}" : "{true_value}")'
```
mr-c, line 375:
```suggestion
                    cwl_command_str = f'$({placeholder_expr} === null ? "{false_value}" : "{true_value}")'
```

PR #121, opened 2022-01-10
translate bcftools_annotate

### LINKED ISSUE #32: Non-input bound declarations are not handled
Currently, non-input bound declarations (which could later be used in the requirements section) are not translated by the convertor and only issues a warning when encountered. 

A WDL workflow containing non-input bound declarations can be found here. (https://github.com/broadinstitute/warp/blob/develop/tasks/broad/Qc.wdl#L26)

Th3nn3ss:
Translate the  bcftools_annotate.wdl file to it's equivalent bcftools_annotate.cwl. 

Fixes #32 

- [x]  Handle inputs with expressions as values. This was achieved by treating the input as optional with a null
- [x]  Handle static values that are not given within inputs. This was achieved by using javascript expressions as a placeholder for that value
- [x]  Characterize the WDL.Expr.IfThenElse expression and handle it's translation to cwl requirements.
- [x] Fix issue with quoted strings used interchangeably in by miniwdl. 
- [x] The class now has two new attributes, `non_static_values` (a set that keeps track of all the values that are not static so as to check if a variable is static in constant time), `optional_cwl_null` (a set that keeps track of inputs that had expressions as values and as such had to be treated as optional)
- [x] Next is to conduct tests


wdl2cwl/main_oop.py
```diff
@@ -205,20 +252,41 @@ def get_expr_apply(self, wdl_apply_expr: WDL.Expr.Apply) -> str:
...
+            elif len(arguments) == 2:
+                operand, regex_str = arguments
+                is_file = isinstance(operand.type, WDL.Type.File)
+                operand = self.get_expr_name(operand.expr)  # type: ignore
+                regex_str = self.get_wdl_literal(regex_str.literal)  # type: ignore
+                regex_str = regex_str[1:]
+                return (
+                    f"{operand}.basename.replace('{regex_str}$', '') "
+                    if is_file
+                    else f"{operand}.split('/').reverse()[0].replace('\.{regex_str}$', '') "
```
mr-c, line 288:
```suggestion
                    else f"{operand}.split('/').reverse()[0].replace("+r"'\."+f"{regex_str}$', '') "
```

wdl2cwl/main_oop.py
```diff
@@ -205,20 +253,41 @@ def get_expr_apply(self, wdl_apply_expr: WDL.Expr.Apply) -> str:
...
+                    f"{only_operand}.basename"
+                    if is_file
+                    else f"{only_operand}.split('/').reverse()[0]"
+                )
+            elif len(arguments) == 2:
+                operand, suffix = arguments
+                is_file = isinstance(operand.type, WDL.Type.File)
+                operand = self.get_expr_name(operand.expr)  # type: ignore
+                suffix_str = self.get_wdl_literal(suffix.literal)  # type: ignore
+                regex_str = suffix_str
```
mr-c, line 285: What happened to the `re.escape`?

Th3nn3ss: Since we are using "/". Javascript has a different escape character I guess.

wdl2cwl/main_oop.py
```diff
@@ -562,8 +649,8 @@ def main() -> None:
         with open(args.output, "w") as result:
             result.write(str(Converter.load_wdl_tree(args.workflow)))
 
-    # Converter.load_wdl_tree("wdl2cwl/tests/wdl_files/bowtie_1.wdl")
-    # Converter.load_wdl_tree("wdl2cwl/tests/wdl_files/bcftools_stats.wdl")
+    Converter.load_wdl_tree("wdl2cwl/tests/wdl_files/bowtie_1.wdl")
+    Converter.load_wdl_tree("wdl2cwl/tests/wdl_files/bcftools_stats.wdl")
```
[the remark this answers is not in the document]
Th3nn3ss: Sorry about that. I leave them commented in case I need it during debugging. Trying to understand how objects are represented as compared to the file am currently working on.

wdl2cwl/main_oop.py
mr-c: Those slashes are JS syntax that inside them is a regular expression. We still need to escape the contents. Otherwise any periods with be interpreted as "any character" instead of "must be the period character", for example

Th3nn3ss: you are totally right. I just tested it. Awesome.

wdl2cwl/main_oop.py
```diff
@@ -511,6 +568,20 @@ def get_cwl_inputs(
 
         return inputs
 
+    def get_input_type(self, input_type: WDL.Tree.Decl) -> str:
+        """Extract the type of any WDL input."""
```
mr-c, line 572:
```suggestion
        """Determine the CWL type for a WDL input declaration."""
```

PR #123, opened 2022-01-13
handle all validateOptimus files 

Th3nn3ss:
Translate all validateOptimus files and return a default of 1GB (1024) for outdirMin in all cwl files (future implementations will handle more derived values)
Also. Return appropriate strings for glob in cwl output derived from wdl outputs with string literals.

### DIFF
--- a/wdl2cwl/main_oop.py
+++ b/wdl2cwl/main_oop.py
@@ -2,7 +2,6 @@
 import os
 import re
 from typing import List, Union, Optional, Callable, cast, Any, Set
-from typing_extensions import runtime
 import WDL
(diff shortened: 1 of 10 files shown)


wdl2cwl/main_oop.py
```diff
@@ -2,6 +2,7 @@
 import os
 import re
 from typing import List, Union, Optional, Callable, cast, Any, Set
+from typing_extensions import runtime
```
mr-c, line 5: It is back! 😂

mr-c (changes requested):

mr-c (approved):

PR #125, opened 2022-01-18
Use a convert function like the former main file

kinow:
It is strange to have to import a class to call a static method, which will then create an instance of the class and call a function.

This PR exports a `convert(str) -> str` function, which has the exact same signature from the existing `main.py`. The `convert` function will then create the object and call its function passing any parameters if necessary.

I think this will be simpler from a user POV.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,3 +12,7 @@ include_trailing_comma = True
 force_grid_wrap = 0
 use_parentheses = True
 line_length = 88
+
+[options.entry_points]
+console_scripts=
+    wdl2cwl = wdl2cwl.main_oop:main
--- a/wdl2cwl/main_oop.py
+++ b/wdl2cwl/main_oop.py
@@ -1,12 +1,11 @@
 """Main entrypoint for WDL2CWL."""
 import os
 import re
-from typing import List, Union, Optional, Callable, cast, Any, Set
+from typing import List, Union, Optional, Any, Set
 import WDL
 import cwl_utils.parser.cwl_v1_2 as cwl
 import regex  # type: ignore
 
-from io import StringIO
 import textwrap
 import sys
 import argparse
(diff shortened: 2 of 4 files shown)


wdl2cwl/main_oop.py
```diff
@@ -718,7 +718,7 @@ def main() -> None:
     # write to a file in oop_cwl_files
     if args.output:
         with open(args.output, "w") as result:
-            result.write(str(Converter.load_wdl_tree(args.workflow)))
+            result.write(convert(args.workflow))
```
mr-c, line 721: So if `--output`/`-o` is missing, we don't do any conversion? :-)

mr-c (changes requested):
In general, yes; thanks! Can the parts with no test coverage either have a test added or be removed?

wdl2cwl/main_oop.py
kinow: I had noticed this in the previous `main.py` too. Good opportunity to fix that. Should we make he `-o` mandatory, or by default print to stdout?

mr-c: not mandatory, no, but remove this https://github.com/kinow/wdl-cwl-translator/blob/7653428821859abd1cdcc03155e86a7e06f633a2/wdl2cwl/main_oop.py#L123

or maybe refactor so that a handle is passed to that function and the `main` function decides if it is a file handle or `sys.stdout`?

setup.cfg
```diff
@@ -12,3 +12,7 @@ include_trailing_comma = True
 force_grid_wrap = 0
 use_parentheses = True
 line_length = 88
+
+[options.entry_points]
+console_scripts=
+    wdl-cwl-translator = wdl2cwl.main_oop:main
```
kinow, line 18: I thought it would be useful to have the command in the `$PATH` of the virtual environment to test workflows? I was calling the `main_oop` directly or editing tests. But I think ultimately we want to give a tool to users? WDYT @mr-c ?

wdl2cwl/main_oop.py
```diff
@@ -24,32 +23,34 @@
 )
 
 
-class Converter:
-    """Object that handles WDL Workflows and task conversion to CWL."""
+def convert(doc: str) -> cwl.CommandLineTool:
```
kinow, line 26: @mr-c this function was doing the conversion and handling serialization too. So I would have to pass the output parameter to choose between `stdout` or file output... I thought it simpler to make it do just the conversion (as per its own name) and leave the serialization up to the user. Now it returns a CWL object.

kinow (commented):
Had a go at moving the code around while allowing users to specify the output destination. Not sure if this is the best design, but happy to take any feedback :+1:

kinow:
Fixing CI...

setup.cfg
mr-c: Can you name the script `wdl2cwl`?

mr-c (approved):

mr-c (approved):

PR #126, opened 2022-01-18
various improvements

kinow:
Reviewing the Python code in `main_oop.py`, seeing where things could be changed/improved.

There are a few functions/things that I didn't understand in the first read, so will pause and continue later.

- [x] raise Exception for missing `else` branch in `if` statement
- [x] review spelling of functions and docs
- [x] simplify functions, move common code to the same functions


wdl2cwl/main_oop.py
```diff
@@ -66,76 +66,72 @@ def load_wdl_task(self, obj: WDL.Tree.Task) -> str:
...
-            else 1024
-        )
-        requirements.append(
-            cwl.ResourceRequirement(
-                coresMin=cpu_requirement,
-                ramMin=memory_requirement,
-                outdirMin=outdir_requirement,
-            )
-        )
+        requirements = self.get_cwl_requirements(obj.command, obj.runtime)
```
kinow, line 69: Realized most of `load_wdl_task` was doing was creating requirements objects/lists. So moved that to its own function.

wdl2cwl/main_oop.py
```diff
@@ -66,76 +66,72 @@ def load_wdl_task(self, obj: WDL.Tree.Task) -> str:
...
-        return 1
+        raise NotImplementedError()
 
     def get_input(self, input_name: str) -> str:
-        """Produce a consise, valid CWL expr/param reference lookup string for a given input name."""
+        """Produce a concise, valid CWL expr/param reference lookup string for a given input name."""
         if valid_js_identifier.match(input_name):
             return f"inputs.{input_name}"
-        return f'inputs["{input_name}"]'  # pragma: no cover
+        return f'inputs["{input_name}"]'
```
kinow, line 134: I think this is for coverage? We have a lot of other code that's not covered yet, and causes Codecov to fail (e.g. #125 ). So I thought we could remove these and do a review of coverage prior to the final release?

wdl2cwl/main_oop.py
```diff
@@ -443,29 +439,33 @@ def get_cpu_requirement(self, cpu_runtime: WDL.Expr.Base) -> str:
...
 
     def get_cwl_docker_requirements(
-        self, wdl_docker: Union[WDL.Expr.Get, WDL.Expr.String]
+        self, wdl_docker: WDL.Expr.Base
     ) -> cwl.ProcessRequirement:
         """Translate WDL Runtime Docker requirements to CWL Docker Requirement."""
+        if not isinstance(wdl_docker, (WDL.Expr.Get, WDL.Expr.String)):
+            raise ValueError(
+                f"Unsupported docker runtime type: {type(wdl_docker)}: {wdl_docker}"
+            )
```
kinow, line 448: This was being done before calling the function. But since both check + function are a pair, and must be executed in order (i.e. you first check the instance, and then call the function) these can be merged and have the check inside the function, avoiding possible bugs where we forget to call the check before calling function...

wdl2cwl/main_oop.py
```diff
@@ -66,76 +66,72 @@ def load_wdl_task(self, obj: WDL.Tree.Task) -> str:
...
         scalarstring.walk_tree(cwl_result)
+        result_stream = StringIO()
         yaml.dump(cwl_result, result_stream)
         yaml.dump(cwl_result, sys.stdout)
 
         return result_stream.getvalue()
 
-    def get_outdire_requirement(
+    def get_cwl_requirements(
+        self, command: WDL.Expr.String, runtime: WDL.Dict[str, WDL.Expr.Base]
```
mr-c, line 93:
```suggestion
        self, command: WDL.Expr.String, runtime: Dict[str, WDL.Expr.Base]
```

and add `Dict` to the `from typing import List, Union, Optional, Callable, cast, Any, Set` line at the top :-)

wdl2cwl/main_oop.py
```diff
@@ -66,76 +66,72 @@ def load_wdl_task(self, obj: WDL.Tree.Task) -> str:
...
+    ) -> List[cwl.ProcessRequirement]:
+        """Produce the CWL Requirements list."""
+        docker_requirement = self.get_cwl_docker_requirements(runtime["docker"])
+        cwl_command_str = self.get_cwl_command_requirements(command.parts)
+        cpu_requirement = (
+            self.get_cpu_requirement(runtime["cpu"]) if "cpu" in runtime else None
+        )
+        memory_requirement = (
+            self.get_memory_requirement(runtime["memory"])  # type: ignore
```
mr-c, line 102:
```suggestion
            self.get_memory_requirement(runtime["memory"])
```

wdl2cwl/main_oop.py
```diff
@@ -66,76 +66,72 @@ def load_wdl_task(self, obj: WDL.Tree.Task) -> str:
...
+            if "memory" in runtime
+            else None
+        )
+        outdir_requirement = (
+            self.get_outdir_requirement(runtime["disks"])  # type: ignore
```
mr-c, line 107:
```suggestion
            self.get_outdir_requirement(runtime["disks"])
```

mr-c (approved):

mr-c:
Thank you @kinow !

PR #133, opened 2022-01-19
Update minimum Python in readme

kinow:
Small PR before calling it a day :sleeping_bed: 

I think we recently changed the minimum Python version to 3.7.


mr-c:
Thanks!

PR #134, opened 2022-01-19
fix indentation in command string

### LINKED ISSUE #130: Make sure no runtime errors if we have more than one `$(` in the WDL command
While reviewing the code today, part of #126 , I noted this line: https://github.com/common-workflow-lab/wdl-cwl-translator/blob/main/wdl2cwl/main_oop.py#L578

It doesn't handle if you have more than `$(` in the string. Looking at the spec, it doesn't say anything about how many `$(` we can have in a WDL command string. But this workflow has multiple: https://github.com/broadinstitute/viral-pipelines/blob/fa070138543df3b6353a17e1821419eb5bd6e87c/pipes/WDL/tasks/tasks_nextstrain.wdl#L486-L499

Th3nn3ss:
Fixes #130

### DIFF
--- a/wdl2cwl/main_oop.py
+++ b/wdl2cwl/main_oop.py
@@ -498,7 +498,7 @@ class Converter:
         command_str: str = ""
         for wdl_command in wdl_commands:
             if isinstance(wdl_command, str):
-                command_str += self.translate_wdl_str(wdl_command)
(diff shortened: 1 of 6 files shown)


mr-c (changes requested):
Great!

mr-c (approved):

PR #135, opened 2022-01-19
delete and replace old files. Also delete references to oop

Th3nn3ss:


wdl2cwl/tests/test_cwl.py
```diff
@@ -1,137 +1,67 @@
...
+    """Get file."""
     return os.path.join(os.path.dirname(__file__), path)
 
 
-def test_meta(capsys: pytest.CaptureFixture[str]) -> None:
-    """Test meta warning."""
-    wdl.convert(get_file("wdl_files/UmiCorrection.wdl"))
-    assert "----WARNING: SKIPPING META----" in capsys.readouterr().err
-
-
```
mr-c, line 18: No more warning?

wdl2cwl/main.py
```diff
@@ -36,784 +21,773 @@
     r"private|continue|debugger|function|arguments|interface|protected|implements|"
     r"instanceof)$)(?:[$_\p{ID_Start}])(?:[$_\u200C\u200D\p{ID_Continue}])*$"
 )
-# ^^ is a combination of https://github.com/tc39/proposal-regexp-unicode-property-escapes#other-examples
-# and regex at the bottom of https://stackoverflow.com/a/9392578
-# double checked against https://262.ecma-international.org/5.1/#sec-7.6
-# eval is not on the official list of reserved words, but it is a built-in function
```
mr-c, line 42: Can we keep this comment?

wdl2cwl/main.py
```diff
@@ -36,784 +21,773 @@
...
-                        append_str = f"$({inputs(sub_str)}.path)"
-                    else:
-                        append_str = "$(inputs." + sub_str + ")"
+    #     elif isinstance(obj, WDL.Workflow):
+    #         return self.load_wdl_workflow(obj)
 
-                new_command = new_command + append_str
+    # def load_wdl_workflow(self, obj: WDL.Workflow):
+    #     print(f"Workflow {obj.name} loaded")
+    #     pass
```
mr-c, line 64: Can this be deleted?

wdl2cwl/tests/test_cwl.py
Th3nn3ss: We don't have the warning int the new main.py

wdl2cwl/main.py
```diff
@@ -36,784 +21,788 @@
...
-                    append_str_common = (
-                        append_str_sub
-                        + ".replace("
-                        + temp[1]
-                        + ","
-                        + temp[2][:-1]
-                        + ")"
-                    )
+    if doc_tree.workflow:
+        return parser.load_wdl_objects(doc_tree.workflow)
```
mr-c, line 39: Unused, so delete or comment out?

wdl2cwl/main.py
```diff
@@ -36,784 +21,788 @@
...
+            raise Exception(f"{type(expr)} has no attribute 'parent'")
+        # if the literal expr is used inside WDL.Expr.Apply
+        # the literal value is what's needed
+        if isinstance(expr.parent, WDL.Expr.Apply):  # type: ignore
+            return self.get_wdl_literal(expr.literal)  # type: ignore
+        # if it is a WDL.Expr.Ident
+        # the parent expression name is needed
+        parent_name = expr.parent.name  # type: ignore
+        parent_name = self.get_input(parent_name)
+        return (  # type: ignore
```
mr-c, line 285: This code is never called, is it still needed?

wdl2cwl/main.py
```diff
@@ -36,784 +21,788 @@
...
+        elif function_name == "_interpolation_add":
+            arg_value, arg_name = arguments
+            if isinstance(arg_name, WDL.Expr.String) and isinstance(
+                arg_value, WDL.Expr.Apply
+            ):
+                arg_name = self.get_expr(arg_name)  # type: ignore
+                arg_value = self.get_expr_apply(arg_value)  # type: ignore
+                return self.get_pseudo_interpolation_add(arg_value, arg_name)  # type: ignore
+            elif isinstance(arg_value, WDL.Expr.Get):
+                arg_name, arg_value = arg_value, arg_name
```
mr-c, line 383: This line is never called; is it still needed?

wdl2cwl/main.py
```diff
@@ -36,784 +21,788 @@
...
+                true_value = options["true"]
+                false_value = options["false"]
+                if type(true_value) == str:
+                    true_str = (
+                        f'"{true_value}"'
+                        if '"' not in true_value
+                        else f"'{true_value}'"
+                    )
+                else:
+                    true_str = true_value
```
mr-c, line 544: This line is not called. Can you create a copy of an existing WDL file and modify it so that this code path is tested?

wdl2cwl/main.py
```diff
@@ -36,784 +21,788 @@
...
+                if type(false_value) == str:
+                    false_str = (
+                        f'"{false_value}"'
+                        if '"' not in false_value
+                        else f"'{false_value}'"
+                    )
+                else:
+                    false_str = false_value
```
mr-c, line 552: Same request here

mr-c (changes requested):

wdl2cwl/main.py
```diff
@@ -36,784 +21,788 @@
...
+            return literal * 60  # type: ignore
+        time_minutes_str = self.get_expr(time_minutes)
+        return f"$({time_minutes_str} * 60)"
+
+    def get_outdir_requirement(
+        self, outdir: Union[WDL.Expr.Get, WDL.Expr.Apply]
+    ) -> int:
+        """Produce the memory requirement for the output directory from WDL runtime disks."""
+        # This is yet to be implemented. After Feature Parity.
+        return int(self.get_wdl_literal(outdir.literal)) * 1024  # type: ignore
```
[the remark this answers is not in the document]
Th3nn3ss: for CollectQualityYeildMetrics.wdl and CollectReadgroupBamQaulity.wdl

wdl2cwl/main.py
```diff
@@ -36,784 +21,788 @@
...
+        elif isinstance(wdl_expr, WDL.Expr.Get):
+            return self.get_expr_get(wdl_expr)
+        elif isinstance(wdl_expr, WDL.Expr.IfThenElse):
+            return self.get_expr_ifthenelse(wdl_expr)
+        elif isinstance(wdl_expr, WDL.Expr.Placeholder):
+            return self.translate_wdl_placeholder(wdl_expr)
+        elif isinstance(wdl_expr, WDL.Expr.String):
+            return self.get_expr_string(wdl_expr)
+        elif isinstance(wdl_expr, WDL.Expr.Ident):
+            return self.get_expr_ident(wdl_expr)
```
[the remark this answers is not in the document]
Th3nn3ss: most WDL.Expr.Ident are wrapped in WDL.Expr.Get and when member is None in WDL.Expr.Get it then passes it's expression to self.get_expr_ident( )

wdl2cwl/main.py
Th3nn3ss: It is used. It was used for the file you edited to make it possible for tests. bcftools_stats.wdl line 75

wdl2cwl/main.py
Th3nn3ss: I just confirmed from the wdl spec that both true and false always have a string literal. I will remove the condition check

wdl2cwl/main.py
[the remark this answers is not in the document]
mr-c: Okay, they weren't getting tested. I fixed that and now theses lines are being run the by tests.

wdl2cwl/main.py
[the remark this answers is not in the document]
mr-c: Ah. Well, this line isn't being run by any of the tests...

wdl2cwl/main.py
mr-c: While `bcftools_stats` is being tested, it no longer triggers this line. Please investigate. I'd like to get the code coverage back up to at least where it was prior.

wdl2cwl/main.py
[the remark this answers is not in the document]
Th3nn3ss: Yeah. That's very possible because of the wrapping in WDL.Expr.Get I will remove it.

wdl2cwl/main.py
```diff
@@ -36,784 +21,766 @@
...
+            WDL.Expr.Array,
+        ],
+    ) -> str:
+        """Translate WDL Boolean, Int or Float Expression."""
+        if expr is None or not hasattr(expr, "parent"):
+            raise Exception(f"{type(expr)} has no attribute 'parent'")
+        # if the literal expr is used inside WDL.Expr.Apply
+        # the literal value is what's needed
+        if isinstance(expr.parent, WDL.Expr.Apply):  # type: ignore
+            return self.get_wdl_literal(expr.literal)  # type: ignore
```
mr-c, line 277: And if the parent isn't a `WDL.Expr.Apply`, what then? (This is caught by `mypy`: https://github.com/common-workflow-lab/wdl-cwl-translator/runs/4885522368?check_suite_focus=true#step:7:17 )

wdl2cwl/main.py
```diff
@@ -36,784 +21,766 @@
...
-            )
-        )
+    def get_wdl_literal(
+        self, wdl_expr: Union[WDL.Expr.Int, WDL.Expr.Float, WDL.Expr.Boolean]
+    ) -> Any:
+        """Extract Literal value from WDL expr."""
+        if wdl_expr is None or not hasattr(wdl_expr, "value"):
+            raise Exception(f"{type(wdl_expr)} has not attribute 'value'")
+        value = wdl_expr.value
+        return value
```
mr-c, line 592: All of the possible inputs have a `value` attribute; why do we check for it? Can't this function be removed and instead accessing the  `.value` attribute directly?

wdl2cwl/main.py
```diff
@@ -36,784 +21,788 @@
...
+        ram_min = self.get_expr_string(memory_runtime)[1:-1]
+        unit = re.search(r"[a-zA-Z]+", ram_min).group()  # type: ignore
+        value = float(ram_min.split(unit)[0])
+
+        if unit == "KiB":
+            memory = value / 1024
+        elif unit == "MiB":
+            memory = value
+        elif unit == "GiB":
+            memory = value * 1024
```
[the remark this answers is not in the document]
kinow: +1

@Th3nn3ss for this one, my approach would be to

1. write a simple pytest unit test
2. since there is an `if` block being tested, you will probably need several test cases to cover all the `if` branches, so I would start with a parametrized test case, e.g. https://github.com/common-workflow-lab/wdl-cwl-translator/blob/44410610d35849a2ccf30652fe302bb5c4bbf032/wdl2cwl/tests/test_cwl.py#L122-L137
3. instead of calling `wdl.convert`, my test would create an instance of the Converter, and then call `get_memory_literal`

On my IDE, I would then run that test separately, sometimes debugging to make sure it is covering the branches the way I expect it. Once all tests are written and I am happy with, then I would run the tests with coverage, and use `make` and/or `tox`. Then update the pull request.

We can have a one-on-one chat, maybe using the VSCode to share your screen, and try to work on this one together if you'd like.

A test like this one should take between 10 minutes and 1 hour max I think. Maybe a little more if it is too complicated to create `WDL.Expr.String` objects programmatically. If you think it will take much longer than that, this one-on-one session can help you avoid getting stuck :slightly_smiling_face: Just let me know :+1:

wdl2cwl/main.py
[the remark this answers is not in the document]
mr-c: Can you go ahead and remove this?

PR #136, opened 2022-01-23
add graph for multiple tasks

### LINKED ISSUE #67: support multiple Task definitions in one file via CWL's $graph
To make a CWL `$graph`: New dictionary with 2 keys: “$graph” (a list of the results of calling “save()” on each CWL document), “cwlVersion” (“v1.2”)

Th3nn3ss:
closes #67

Handle multiple tasks in one file using $graph.

### DIFF
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -1,7 +1,7 @@
 """Main entrypoint for WDL2CWL."""
 import os
 import re
-from typing import List, Union, Optional, Any, Set
+from typing import List, Union, Optional, Any, Set, Dict
 import WDL
 import cwl_utils.parser.cwl_v1_2 as cwl
 import regex  # type: ignore
@@ -28,7 +28,7 @@ valid_js_identifier = regex.compile(
 # eval is not on the official list of reserved words, but it is a built-in function
 
 
-def convert(doc: str) -> cwl.CommandLineTool:
+def convert(doc: str) -> Dict[str, List[str]]:
     """Convert a WDL workflow, reading the file, into a CWL workflow Python object."""
     wdl_path = os.path.relpath(doc)
     doc_tree = WDL.load(wdl_path)
@@ -36,18 +36,18 @@ def convert(doc: str) -> cwl.CommandLineTool:
     parser = Converter()
 
     dict_of_tasks = {"cwlVersion": "v1.2"}
-    dict_of_tasks["$graph"] = []
+    dict_of_tasks["$graph"] = []  # type: ignore
 
     for index, task in enumerate(doc_tree.tasks):
         cwl_task_object = parser.load_wdl_objects(task).save()
         if len(doc_tree.tasks) == 1:
             return cwl_task_object
         if index == 0:
-        # change Id name of the first task
+            # change Id name of the first task
             cwl_task_object["id"] = "main"
-        dict_of_tasks["$graph"].append(cwl_task_object)
+        dict_of_tasks["$graph"].append(cwl_task_object)  # type: ignore
 
-    return dict_of_tasks
+    return dict_of_tasks  # type: ignore
 
 
 class Converter:
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -38,13 +38,10 @@ def convert(doc: str) -> Dict[str, List[str]]:
     dict_of_tasks = {"cwlVersion": "v1.2"}
     dict_of_tasks["$graph"] = []  # type: ignore
 
-    for index, task in enumerate(doc_tree.tasks):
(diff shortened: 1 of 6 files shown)


wdl2cwl/tests/wdl_files/rtg_2_tasks.wdl
```diff
@@ -0,0 +1,183 @@
+version 1.0
+
+# Copyright (c) 2020 Leiden University Medical Center
```
mr-c, line 3: Can you add a comment with a permalink to the source of this file?

wdl2cwl/main.py
```diff
@@ -28,18 +28,26 @@
...
+    dict_of_tasks["$graph"] = []  # type: ignore
 
-    return tasks[0]
+    for index, task in enumerate(doc_tree.tasks):
+        cwl_task_object = parser.load_wdl_objects(task).save()
+        if len(doc_tree.tasks) == 1:
+            return cwl_task_object
+        if index == 0:
+            # change Id name of the first task
+            cwl_task_object["id"] = "main"
```
mr-c, line 47: Why?

Th3nn3ss: graph doesn't work without a main id name.

mr-c: With `cwltool` you can execute any item in a `$graph` : `cwltool bcftools.cwl#Sort`

It is true, that if you don't want to include the `#id` part then one of the entries much be called `main`.

Th3nn3ss: What can I do instead of changing the first ID to `main`?

mr-c: Delete this line. No reason to change IDs 😀👍

Th3nn3ss: But it fails when I try to run --validate

mr-c: `cwltool --validate bcftools.cwl#Sort` and `cwltool --validate bcftools.cwl#Annotate`

Th3nn3ss: alright. Cool.

wdl2cwl/main.py
```diff
@@ -28,18 +28,23 @@
 # eval is not on the official list of reserved words, but it is a built-in function
 
 
-def convert(doc: str) -> cwl.CommandLineTool:
+def convert(doc: str) -> Dict[str, List[str]]:
```
kinow, line 31: Isn't that `Dict[str, Any]`, as that's the type returned by `miniwdl` functions used within this function?

wdl2cwl/main.py
```diff
@@ -28,18 +28,23 @@
...
+
     for task in doc_tree.tasks:
-        tasks.append(parser.load_wdl_objects(task))
+        cwl_task_object = parser.load_wdl_objects(task).save()
+        if len(doc_tree.tasks) == 1:
+            return cwl_task_object
+        dict_of_tasks["$graph"].append(cwl_task_object)  # type: ignore
 
-    return tasks[0]
+    return dict_of_tasks  # type: ignore
```
kinow, line 47: @Th3nn3ss I haven't seen you using [list](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions) & [dictionary](https://docs.python.org/3/tutorial/datastructures.html#dictionaries) comprehensions yet. That's something that took me some time getting used to in Python, but it has some performance benefits, besides being really common by Python devs. Here's a variation of this function with list comprehensions, and creating-and-returning the dictionary in one go (while, IMO, still keeping readability - it's important to remember readability for yourself & fellow devs)

```py
def convert(doc: str) -> Dict[str, Any]:
    """Convert a WDL workflow, reading the file, into a CWL workflow Python object."""
    wdl_path = os.path.relpath(doc)
    doc_tree = WDL.load(wdl_path)

    parser = Converter()

    if len(doc_tree.tasks) == 1:
        return parser.load_wdl_objects(doc_tree.tasks[0]).save()
    else:
        return {
            "cwlVersion": "v1.2",
            "$graph": [parser.load_wdl_objects(task).save() for task in doc_tree.tasks],
        }
```

p.s.: list/dict comprehension can also lead to performance [problems](https://realpython.com/list-comprehension-python/#when-not-to-use-a-list-comprehension-in-python), actually, and make the code harder to read & debug, so you don't need to replace every for-loop by a list comprehension :slightly_smiling_face:

kinow: I replaced the return type, and removed the `type: ignore` markers, and `make mypy` passed for me locally with the code above. Maybe changing the return type will remove the ned to ignore types here @Th3nn3ss ? The list comprehension change is optional :+1:

mr-c (approved):
Thanks!

Issue #104, opened 2021-11-06
implement WDL's `ceil()` function


mr-c:
https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#int-floorfloat-int-ceilfloat-and-int-roundfloat

[`Math.ceil(x)`](https://262.ecma-international.org/5.1/#sec-15.8.2.6) in ECMAScript 5.1

example: https://github.com/common-workflow-lab/wdl-cwl-translator/blob/7cd094a27f4e74ba5a3d67dc2aed0fece6766d99/wdl2cwl/tests/wdl_files/hisat2_1.wdl#L53

> ` ceil(threads / 4.0)`

equivalent to the CWL expression snippet `Math.ceil(inputs.threads/ 4.0)`

PR #137, opened 2022-01-24
Size ceil functions

### LINKED ISSUE #103: WDL's `size()` function
https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#float-sizefilearrayfile-string

https://www.commonwl.org/v1.2/CommandLineTool.html#File look for the `size` field which is measured in bytes.

example: https://github.com/common-workflow-lab/wdl-cwl-translator/blob/7cd094a27f4e74ba5a3d67dc2aed0fece6766d99/wdl2cwl/tests/wdl_files/CollectQualityYieldMetrics.wdl#L28

> `size(input_bam, "GiB")`

equivalent to the CWL expression snippet `inputs.input_bam.size / 1024^3`

Th3nn3ss:
closes #103 & #104 

The size function is a WDL.Expr.Apply expression that accepts two arguments, the input file name and the memory unit in Binary. 
The ceil function is a WDL.Expr.Apply expression that accepts only one argument, usually a float, and returns the upper bound of the literal. 

The javascript Math.ceil is used in the retuned cwl_expression.

### DIFF
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -156,8 +156,8 @@ class Converter:
             if isinstance(outdir, WDL.Expr.String):
                 for part in outdir.parts:
                     if isinstance(part, WDL.Expr.Placeholder):
-                        outdir_str = self.get_expr(part)
-                        return outdir_str
+                        outdir_str = self.get_expr(part)[1:]
+                        return f"$({outdir_str}*1024)"
             else:
                 raise Exception(f"Runtime Disks {outdir} are not handled yet.")
         return int(outdir.literal.value) * 1024  # type: ignore
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -152,10 +152,15 @@ class Converter:
         self, outdir: Union[WDL.Expr.Get, WDL.Expr.Apply, WDL.Expr.String]
     ) -> Union[int, str]:
         """Produce the memory requirement for the output directory from WDL runtime disks."""
-        if outdir.literal:
-            return int(outdir.literal.value) * 1024
-        outdir_str = self.get_expr(outdir)
-        return f"$({outdir_str})"
+        if not outdir.literal:
+            if isinstance(outdir, WDL.Expr.String):
+                for part in outdir.parts:
+                    if isinstance(part, WDL.Expr.Placeholder):
+                        outdir_str = self.get_expr(part)
+                        return outdir_str
+            else:
+                raise Exception(f"Runtime Disks {outdir} are not handled yet.")
+        return int(outdir.literal.value) * 1024  # type: ignore
 
     def get_input(self, input_name: str) -> str:
         """Produce a consise, valid CWL expr/param reference lookup string for a given input name."""
@@ -417,7 +422,7 @@ class Converter:
         elif function_name == "ceil":
             only_arg = arguments[0]
             only_arg_expr = self.get_expr(only_arg)
-            return f"Math.ceil({only_arg_expr})"
+            return f"Math.ceil({only_arg_expr}) "
         elif function_name == "_mul":
             left_operand, right_operand = arguments
             if left_operand.literal:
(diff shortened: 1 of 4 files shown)


wdl2cwl/main.py
```diff
@@ -284,10 +286,7 @@ def get_expr_string(self, wdl_expr_string: WDL.Expr.String) -> str:
         string = ""
         parts = wdl_expr_string.parts
         for index, part in enumerate(parts[1:-1], start=1):
-            if isinstance(
-                part,
-                (WDL.Expr.Placeholder, WDL.Expr.Apply, WDL.Expr.Get, WDL.Expr.Ident),
-            ):
+            if isinstance(part, WDL.Expr.Placeholder):
```
Th3nn3ss, line 289: only placeholders in strings have been observed. No other expression seems possible to be embedded in a string.

wdl2cwl/tests/cwl_files/TrimAdapters_without_comment.cwl
```diff
@@ -0,0 +1,98 @@
...
+        else if(unit==="TiB") memory = value*1024*1024;
+        else if(unit==="B") memory = value/(1024*1024);
+        else if(unit==="KB" || unit==="K") memory = (value*1000)/(1024*1024);
+        else if(unit==="MB" || unit==="M") memory = (value*(1000*1000))/(1024*1024);
+        else if(unit==="GB" || unit==="G") memory = (value*(1000*1000*1000))/(1024*1024);
+        else if(unit==="TB" || unit==="T") memory = (value*(1000*1000*1000*1000))/(1024*1024);
+        return parseInt(memory);
+        }
+    outdirMin: $('local-disk ' + $(Math.ceil(2*(inputs.fastq1_input_files.size / 1024^3
+        + inputs.fastq2_input_files.size / 1024^3)) + 10) + ' HDD')
```
mr-c, line 94:
```suggestion
    outdirMin: $(Math.ceil(2*(inputs.fastq1_input_files.size / 1024^3
        + inputs.fastq2_input_files.size / 1024^3)) + 10)
```

https://www.commonwl.org/v1.2/CommandLineTool.html#ResourceRequirement  ; when we don't say explicitly, then an `expression` can return any of the other listed types: `long` or `float`, but not `string`. In fact, the concept of `local-disk` and `HDD` does not exist in CWL

Additionally, you can't nest one `$()` expression in another `$()` with duplicate `$(`; remove the inner`$( ... )`

Though, is this the right calculation? The source `disks` declaration does not include any units, so the default according to https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#disks is `GiB`

wdl2cwl/main.py
```diff
@@ -149,10 +149,17 @@ def get_time_minutes_requirement(
...
+        self, outdir: Union[WDL.Expr.Get, WDL.Expr.Apply, WDL.Expr.String]
+    ) -> Union[int, str]:
         """Produce the memory requirement for the output directory from WDL runtime disks."""
-        # This is yet to be implemented. After Feature Parity.
+        if not outdir.literal:
+            if isinstance(outdir, WDL.Expr.String):
+                for part in outdir.parts:
+                    if isinstance(part, WDL.Expr.Placeholder):
+                        outdir_str = self.get_expr(part)[1:]
+                        return f"$({outdir_str}*1024)"
```
mr-c, line 160: What if the WDL `disks` entry has used non-default units like ` disks: "local-disk ${disk} MiB" ?

wdl2cwl/main.py
```diff
@@ -409,13 +413,40 @@ def get_expr_apply(self, wdl_apply_expr: WDL.Expr.Apply) -> str:  # type: ignore
...
+                right_operand_value = self.get_expr(right_operand)
+            return f"{left_operand_value}*({right_operand_value}) "
+        elif function_name == "size":
+            left_operand, right_operand = arguments
+            input_file_name = self.get_expr_name(left_operand.expr)  # type:ignore
+            size_unit = self.get_expr(right_operand)[1:-1]
+            if size_unit == "Gi":
+                unit_value = 3
+            else:
+                raise ValueError(f"Size unit {size_unit} not handle yet")
```
mr-c, line 448: I get this, but it would be nicer to handle all the WDL units now https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#units-of-storage

mr-c (changes requested):
If this PR is to fix #103 and #104 then I would expect both example listed in those issues to be demonstrated here as well

mr-c:
superseded by #150

PR #151, opened 2022-02-02
Try a different module/package structure

kinow:
It started over the week/holiday when I had a go at trying to write the conversion for WDL stdlib's `select_all` function. It didn't work for [other reasons](https://github.com/chanzuckerberg/miniwdl/issues/541), but I noticed that the `main.py` keeps getting longer and longer.

So I thought we could probably reduce it a little by starting to use more modules/packages.

After reading the `miniwdl` code, I thought we could have a structure similar to their structure, but also thinking about CWL.

With that in mind, I thought about an `wdl2cwl.expr` module with the WDL Expr statements (conditionals, arithmetic, etc). And then remove the `if` statement from the `WDL.Expr.Apply` for WDL Stdlib Functions, and move that code to a separate module as well.

That resulted in `wdl2cwl.functions`. Furthermore, instead of using an `if/elif's/else` statement, we can simply check if the module has a function with the same name (coding by convention) and use it if so, otherwise raise an error as before.

While that would reduce the code, not sure if that's the best/right call for now. So happy to get any feedback. I almost finished this change, but ran out of :coffee: (head was literally bobbing while I was writing the code :sleeping: , will review in the morning :sleeping_bed: )

Cheers
-Bruno


wdl2cwl/main.py
```diff
@@ -113,52 +96,11 @@ def get_cwl_docker_requirements(
...
-
-
 class Converter:
     """Object that handles WDL Workflows and task conversion to CWL."""
 
     def __init__(self) -> None:
-        """Initialize the sets used by the object and prevent inconsistent behaviours."""
-        self.non_static_values: Set[str] = set()
-        self.optional_cwl_null: Set[str] = set()
+        self.ctx = ConversionContext()
```
kinow, line 103: Note to self: Probably good idea to reset the context (clear set) when the file was written to the output stream (i.e. when conversion has finished), to avoid stale state...

kinow: And made it into a dataclass so I could pass it around as a single variable.

wdl2cwl/expr.py
```diff
@@ -0,0 +1,288 @@
...
+    treat_as_optional = wdl_apply_expr.type.optional
+
+    # Call the function if we have it in our wdl2cwl.functions module
+    if hasattr(functions, function_name):
+        return cast(
+            str,
+            getattr(functions, function_name)(
+                arguments, ctx, treat_as_optional=treat_as_optional
+            ),
+        )
```
kinow, line 195: The rest of the code in this module should be very similar (if not exact copy) of the code from `main.py`. This `_apply` function was modified to dynamically call a WDL Stdlib by its name from `wdl2cwl.functions`.

wdl2cwl/functions.py
```diff
@@ -0,0 +1,94 @@
...
+from .expr import (
+    get_expr,
+    get_expr_name,
+    get_expr_apply,
+    get_expr_name_with_is_file_check,
+)
+from .util import ConversionContext, get_input
+
+
+def _add(arguments: List[Base], ctx: ConversionContext, **kwargs: Any) -> str:
```
kinow, line 18: Signature for the functions is:

`fn ( arguments, ctx, **kwargs )` where

- `fn` must match a function from WDL's [stdlib](https://github.com/chanzuckerberg/miniwdl/blob/5d5b7f8e0147278a7a211f13ad863f6586f902b8/WDL/StdLib.py#L306).
- `arguments` is same as before, the list of args for this function
- `**kwargs` whatever else is passed as a dict/param, like that optional flag

wdl2cwl/main.py
```diff
@@ -997,7 +567,7 @@ def get_cwl_task_outputs(
                     )
                 )
             else:
-                glob_expr = self.get_expr(wdl_output)
+                glob_expr = get_expr(wdl_output, self.ctx)
                 glob_str = f"$({glob_expr})"
 
                 if wdl_output.type.optional:
```
kinow, line 573: `main.py` has `1056` lines on `main` branch. This PR reduces it to `627`. So reading & maintaining `main.py` should become a bit easier.

kinow:
I suspect the coverage change is due to changes in line numbers.

kinow:
Happy to keep rebasing/fixing conflicts so we can get @Th3nn3ss 's & @mr-c 's PRs in first :wave:

Th3nn3ss (approved):
I don't fully understand every implementation in this PR but am very excited to see how efficiently it handles dependencies between the python files. I hope to have pythonic powers like this someday. Incredible work.

kinow:
Will update this PR over the next days :muscle:

kinow:
Rebased, updated the code, and I think I got all the latest changes :nerd_face: :sleepy: Going to stop for now, and review it in the morning. Then try to wrap it up updating anything broken in the CI. Or, if nothing broken, then just mark it as ready for review.

kinow:
Done!

Issue #105, opened 2021-11-06
WDL `runtime.disks` with suffixes, WDL expressions


mr-c:
Follow on to #95 

https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#disks

`<mount-point>`s should be accepted and their sizes included, but the mount points themselves won't appear in the CWL output as CWL has no concept of per-mountpoint persistent volumes and space reservation.

Example from the current tests:
https://github.com/broadinstitute/warp/blob/9ea4dc8ad4eca794ed336451fa8b23367132a820/tasks/broad/Qc.wdl#L37
> `disks: "local-disk " + disk_size + " HDD"`

`local-disk` is a Cromwell specific mount point, and so should be ignored.

The `HDD` disk type postfix is [Cromwell specific thing](https://cromwell.readthedocs.io/en/stable/RuntimeAttributes/#disks) and not part of OpenWDL, however we must tolerate it by ignoring it as well as `SDD`or `LOCAL` postfixes.

The correct translation of this example for CWL would be

``` cwl
hints:
  ResourceRequirement:
     outdirMin: $(((Math.ceil(inputs.input_bam.size) / 1024^3) + 20) / 1024)
```

https://www.commonwl.org/v1.2/CommandLineTool.html#ResourceRequirement

Th3nn3ss:
```
outdirMin: $((Math.ceil((function(size_of=0){inputs.input_bam.path.forEach(function(element){
     if (element) {size_of += element.size}})}) / 1024^3)  + 20)/ 1024)
```

PR #154, opened 2022-02-04
Outdir min

Th3nn3ss:
Closes: #105 

support disks with WDL.Expr.String
support disks with WDL.Expr.Apply
support memory with WDL.Expr.Placeholder contained in WDL.Expr.String
support WDL.Expr.Apply with function name "read_int"


wdl2cwl/main.py
```diff
@@ -95,13 +95,6 @@ def get_mem_in_bytes(unit: str) -> str:
     return mem_in_bytes
 
 
-def get_outdir_requirement(outdir: Union[WDL.Expr.Get, WDL.Expr.Apply]) -> int:
-    """Produce the memory requirement for the output directory from WDL runtime disks."""
-    # This is yet to be implemented. After Feature Parity.
-    with WDLSourceLine(outdir, ConversionException):
-        return int(outdir.literal.value) * 1024  # type: ignore
-
-
```
Th3nn3ss, line 104: This needed to be in the class.

wdl2cwl/main.py
```diff
@@ -151,7 +144,7 @@ def get_literal_name(
     # if the literal expr is used inside WDL.Expr.Apply
     # the literal value is what's needed
     parent = expr.parent  # type: ignore[union-attr]
-    if isinstance(parent, (WDL.Expr.Apply, WDL.Expr.IfThenElse)):
+    if isinstance(parent, (WDL.Expr.Apply, WDL.Expr.IfThenElse, WDL.Expr.Array)):
```
Th3nn3ss, line 147: For literals that are declared in an Array. This is useful when all the elements in an array are been iterated to be reproduced in a string joined by a coma.

wdl2cwl/main.py
```diff
@@ -351,7 +344,7 @@ def get_cwl_requirements(self, obj: WDL.Tree.Task) -> List[cwl.ProcessRequiremen
             memory_requirement = None
         if "disks" in runtime:
             with WDLSourceLine(runtime["memory"], ConversionException):
-                outdir_requirement = get_outdir_requirement(
+                outdir_requirement = self.get_outdir_requirement(
```
Th3nn3ss, line 347: Since the get_outdir_requirement( ) is now a method and not a function outside the class

wdl2cwl/main.py
```diff
@@ -394,6 +387,10 @@ def get_memory_requirement(
             if isinstance(memory_runtime, WDL.Expr.String):
                 ram_min_literal = self.get_memory_literal(memory_runtime)
                 return ram_min_literal
+            elif isinstance(memory_runtime, WDL.Expr.Apply):
+                expr, unit = memory_runtime.arguments
```
Th3nn3ss, line 391: For instances where the memory is concatenated with a function expression to form a string. Like in the case of `memory: select_first([machine_mem_gb, 13]) + " GB"`

wdl2cwl/main.py
```diff
@@ -419,6 +416,31 @@ def get_memory_literal(self, memory_runtime: WDL.Expr.String) -> float:
...
         return memory
 
+    def get_outdir_requirement(
+        self, outdir: Union[WDL.Expr.Get, WDL.Expr.Apply]
+    ) -> int:
+        """Produce the memory requirement for the output directory from WDL runtime disks."""
+        with WDLSourceLine(outdir, ConversionException):
+            if isinstance(outdir, (WDL.Expr.Apply, WDL.Expr.String)):
+                list_object = (
+                    outdir.arguments if hasattr(outdir, "arguments") else outdir.parts  # type: ignore
```
Th3nn3ss, line 426: If outdir is an expression WDL.Expr.String then it has an iterable object called parts else if it's a WDL.Expr.Apply it has an iterable object called arguments.

wdl2cwl/main.py
```diff
@@ -419,6 +416,31 @@ def get_memory_literal(self, memory_runtime: WDL.Expr.String) -> float:
...
+                )
+                for obj in list_object:
```
Th3nn3ss, line 428: This loop ensures that we are not concerned about strings like "local" and "SDD" or "HDD"

wdl2cwl/tests/cwl_files/validateOptimus_with_disks.cwl
```diff
@@ -0,0 +1,68 @@
...
+            fi
+  - class: InlineJavascriptRequirement
+  - class: NetworkAccess
+    networkAccess: true
+  - class: ResourceRequirement
+    coresMin: 1
+    ramMin: 953.67431640625
+    outdirMin: $((Math.ceil((function(size_of=0){inputs.cell_metrics.path.forEach(function(element){
+        if (element) {size_of += element.size}})}) / 1000^3 + (function(size_of=0){inputs.gene_metrics.path.forEach(function(element){
+        if (element) {size_of += element.size}})}) / 1000^3*1.1) ) / 1024)
```
mr-c, line 64:
```suggestion
    outdirMin: $((Math.ceil(((inputs.cell_metrics.size / 1000^3) + (inputs.gene_metrics.size / 1000^3))*1.1) ) / 1024)
```
When we know the input is a single file and not an array, we should do a simpler size calculation. Extra parenthesis are to prevent order of operation errors

wdl2cwl/main.py
```diff
@@ -419,6 +416,31 @@ def get_memory_literal(self, memory_runtime: WDL.Expr.String) -> float:
...
+                    if isinstance(
+                        obj, (WDL.Expr.Get, WDL.Expr.Apply, WDL.Expr.Placeholder)
+                    ):
+                        return self.get_outdir_requirement(obj)  # type: ignore
+            elif isinstance(outdir, (WDL.Expr.Get, WDL.Expr.Placeholder)):
+                expr = self.get_expr(outdir)
+                expr_str = expr
+                if isinstance(outdir, (WDL.Expr.Placeholder)):
```
Th3nn3ss, line 436: if it's a placeholder. the tralslate_wdl_placeholder function will return with the string `$(expr)`. This removes the first two and last chars of the string

wdl2cwl/main.py
```diff
@@ -633,7 +655,7 @@ def get_expr_apply(self, wdl_apply_expr: WDL.Expr.Apply) -> str:
             return self.get_expr(only_arg)
         elif function_name == "select_first":
             array_obj = arguments[0]
-            array_items = [self.get_expr(item) for item in array_obj.items]  # type: ignore
+            array_items = [str(self.get_expr(item)) for item in array_obj.items]  # type: ignore
```
Th3nn3ss, line 658: as the elements of an array might not be strings this converts every item to a string

wdl2cwl/main.py
```diff
@@ -678,6 +700,8 @@ def get_expr_apply(self, wdl_apply_expr: WDL.Expr.Apply) -> str:
                 + "}})}"
                 + f") / {unit_value}"
             )
+        elif function_name == "read_int":
```
Th3nn3ss, line 703: This handles the "read_int" function.

wdl2cwl/tests/cwl_files/validateOptimus_with_disks.cwl
Th3nn3ss: It's proving hard to determine files with just one element because of struct.

wdl2cwl/tests/wdl_files/BuildCembaReferences.wdl
```diff
@@ -79,15 +81,15 @@ task BuildBisulfiteReferences {
...
+    python3 /build_bisulfite_references.py \
       --input-fasta ~{fasta_input} \
       --forward-convert-out ~{fwd_converted_reference_fasta_output_name} \
       --reverse-convert-out ~{rev_converted_reference_fasta_output_name}
   >>>
 
   runtime {
     docker: "quay.io/broadinstitute/bisulfite-references:1.0"
-    # disks: "local-disk " + ceil(3.5 * (if input_size < 1 then 1 else input_size)) + " HDD"
+    #disks: "local-disk " + ceil(3.5 * (if input_size < 1 then 1 else input_size)) + " HDD"
```
mr-c, line 92: If I uncomment this I get `BuildCembaReferences.wdl:92:33: 'NoneType' object has no attribute 'value'` Can you investigate @Th3nn3ss ?

Th3nn3ss: For this, are we still using `$(expr / 1024)`

wdl2cwl/main.py
```diff
@@ -139,25 +132,6 @@ def get_cwl_docker_requirements(
     return cwl.DockerRequirement(dockerPull=dockerpull)
 
 
-def get_literal_name(
```
Th3nn3ss, line 142: This function no longer serves it's main purpose since we started tracking non_static inputs.

wdl2cwl/main.py
```diff
@@ -514,6 +526,7 @@ def get_expr_ifthenelse(self, wdl_ifthenelse: WDL.Expr.IfThenElse) -> str:
 
     def get_expr_apply(self, wdl_apply_expr: WDL.Expr.Apply) -> str:
         """Translate WDL Apply Expressions."""
+        single_arg_fn = {"read_string", "read_float", "glob", "read_int"}
```
Th3nn3ss, line 529: a set of single argument functions that return a `self.get_expr(argument[0])`

wdl2cwl/tests/cwl_files/validateOptimus_5.cwl
```diff
@@ -51,7 +51,7 @@ requirements:
   - class: ResourceRequirement
     coresMin: 1
     ramMin: 953.67431640625
-    outdirMin: 1024
+    outdirMin: $((1) * 1024)
```
mr-c, line 54: If there are only static values, can we calculate the result instead of using a CWL expression?

Issue #153, opened 2022-02-02
align_and_count.cwl fails validation


mr-c:
See https://github.com/common-workflow-lab/wdl-cwl-translator/runs/5036889858?check_suite_focus=true#step:7:108

mr-c:
@Th3nn3ss the sources are missing the `reports.` in front

https://github.com/common-workflow-lab/wdl-cwl-translator/blob/9781f7b166165e2b06f5cecad49e58df6e24d0d9/wdl2cwl/tests/cwl_files/align_and_count.cwl#L22

should be `reports.align_and_count/report`

https://github.com/common-workflow-lab/wdl-cwl-translator/blob/9781f7b166165e2b06f5cecad49e58df6e24d0d9/wdl2cwl/tests/cwl_files/align_and_count.cwl#L32

PR #155, opened 2022-02-04
Support array globs

mr-c:
Fixes #53

Also support WDL's `size()` with only one argument (default units).

PR #156, opened 2022-02-06
Align and count

Th3nn3ss:
Fixes: #153 

Resolves: #152 

The outputSource for workflow outputs are handled in the get_cwl_task_inputs( this should probably be renamed to get_cwl_inputs at this point) and for referees that are Calls, the length of the callee_id list is checked if it is equal to 2. As imported  inputs have a length of 2. The first index is the origin of the input i.e the task being imported. 

The  glob_str is concatenated with the callee_id[0] and returned as outputSource


wdl2cwl/main.py
```diff
@@ -1014,6 +1014,15 @@ def get_cwl_task_outputs(
                     and wdl_output.expr.literal is not None
                 ):
                     glob_str = glob_str[3:-2]
+
+                if isinstance(wdl_output.expr, WDL.Expr.Get) and isinstance(
```
kinow, line 1018: Could it be an else if, part of the if statement above?

kinow (approved):
One comment, but otherwise looks good to me! :tada:

wdl2cwl/main.py
```diff
@@ -223,7 +223,7 @@ def load_wdl_workflow(self, obj: WDL.Tree.Workflow) -> cwl.Workflow:
             cwl.WorkflowOutputParameter(
                 id=f"{wf_name}.{output.id}",
                 type=output.type,
-                outputSource=output.outputBinding.glob.replace(".", "/")[2:-1]
+                outputSource=output.outputBinding.glob
```
mr-c, line 226: See https://github.com/common-workflow-lab/wdl-cwl-translator/pull/155/files#diff-b5a65c2befbbd0b6453805272729b622f41b2be683f8258432e08f4370332265R226 for a different opinion :-)

PR #161, opened 2022-02-07
Add select_all miniwdl Stdlib function

kinow:


wdl2cwl/tests/test_wdl_stdlib.py
```diff
@@ -0,0 +1,35 @@
...
+        Int one
+        Int? two
+      }
+      command {}
+      output {
+        Array[Int] first1 = select_all([one, two])
+      }
+    }
+    """
+    with NamedTemporaryFile(mode="w", encoding="utf-8") as output:
```
kinow, line 25: @mr-c I wrote this one reading the code [^1] & tests [^2] for `select_all` in miniwdl. I didn't have time to search for a workflow to write the functional test, so quickly wrote a unit test. Do you think it'd be a good idea to add tests like this one? Or do you think it'd be better to only use the functional tests with a given WDL file, and another CWL file with the expected result?

[^1]: https://github.com/chanzuckerberg/miniwdl/blob/0ee2bddaea4bd2538cd21bd9b7806a6c81068b2f/WDL/StdLib.py#L769
[^2]: https://github.com/chanzuckerberg/miniwdl/blob/0ee2bddaea4bd2538cd21bd9b7806a6c81068b2f/tests/test_5stdlib.py#L156

wdl2cwl/main.py
```diff
@@ -654,6 +654,11 @@ def get_expr_apply(self, wdl_apply_expr: WDL.Expr.Apply) -> str:
             array_items = [str(self.get_expr(item)) for item in array_obj.items]  # type: ignore
             items_str = ", ".join(array_items)
             return f"[{items_str}].find(element => element !== null) "
+        elif function_name == "select_all":
+            array_obj = cast(WDL.Expr.Array, arguments[0])
+            array_items = [str(self.get_expr(item)) for item in array_obj.items]
+            items_str = ", ".join(array_items)
+            return f"[{items_str}].filter(element => element !== null) "
```
kinow, line 661: Basically copied the `select_first`, but instead of JS's `.find`, used `.filter`.

wdl2cwl/errors.py
```diff
@@ -46,7 +45,7 @@ def __exit__(
 
     def makeLead(self) -> str:
         """Caculate the error message prefix."""
-        pos: SourcePosition = cast(SourcePosition, self.item.pos)
+        pos: WDL.SourcePosition = cast(WDL.SourcePosition, self.item.pos)
```
kinow, line 48: The import used breaks the tests on my IDE when trying to run a test individually. miniwdl exports `SourcePosition` in the main export, so I think it should be OK to use it instead, since it also fixes the IDE issue.

wdl2cwl/tests/test_wdl_stdlib.py
```diff
@@ -0,0 +1,35 @@
...
+        output.write(wdl)
+        output.flush()
+        cwl = convert(output.name)
+        assert cwl["outputs"][0] == {
+            "id": "first1",
+            "type": {"items": "int", "type": "array"},
+            "outputBinding": {
+                "glob": "$([inputs.one, inputs.two].filter(element => element !== null) )"
```
mr-c, line 33: This doesn't make sense; why are we `glob`ing on int values?

It would be better to use standalone WDL and CWL files, so we can also test the resulting CWL files more directly.

mr-c (changes requested):
`select_all` is typically paired with `if` or `scatter` in WDL, so it might be a bit soon to implement this.

Can you make your branch on the main repo? It makes it simpler to collaborate :-)

Maybe consider using https://github.com/griffithlab/analysis-wdls/blob/25cf2010055d79bf7df63ab832363066eba87a26/definitions/subworkflows/merge_svs.wdll as an example? There `select_all` on a source selector for a workflow step could be represented in CWL by using `pickValue: all_non_null`

wdl2cwl/tests/test_wdl_stdlib.py
kinow: >This doesn't make sense; why are we globing on int values?

Does this make more sense @mr-c ?

```wdl
version 1.0
task test_select_all {
  input {
    Int one
    Int? two
  }
  command {}
  output {
    Array[Int] first1 = select_all([one, two, 1, select_first([two, one])])
  }
}
```

```cwl
class: CommandLineTool
id: test_select_all
inputs:
  - id: one
    type: int
  - id: two
    type:
      - int
      - 'null'
outputs:
  - id: first1
    type:
        items: int
        type: array
    outputBinding:
        outputEval: $([inputs.one, inputs.two, 1, [inputs.two, inputs.one].find(function(element)
            { element !== null }) ].filter(function(element) { element !== null })
            )
requirements:
  - class: InitialWorkDirRequirement
    listing:
      - entryname: script.bash
        entry: ''
  - class: InlineJavascriptRequirement
  - class: NetworkAccess
    networkAccess: true
  - class: ResourceRequirement
    outdirMin: 1024
cwlVersion: v1.2
baseCommand:
  - bash
  - script.bash
```
Looks like the string returned by the evaluation of the `WDL.Expr.Apply` is [always used](https://github.com/common-workflow-lab/wdl-cwl-translator/blob/881e646cdc552b1e496fb8833ead0df795382c47/wdl2cwl/main.py#L1111) for glob in the output.

>It would be better to use standalone WDL and CWL files, so we can also test the resulting CWL files more directly.

Will push a commit with the file and some WIP code. Marking the PR as draft :+1: 

Thanks!

wdl2cwl/main.py
```diff
@@ -650,10 +650,15 @@ def get_expr_apply(self, wdl_apply_expr: WDL.Expr.Apply) -> str:
             only_arg = arguments[0]
             return self.get_expr(only_arg)
         elif function_name == "select_first":
-            array_obj = arguments[0]
-            array_items = [str(self.get_expr(item)) for item in array_obj.items]  # type: ignore
+            array_obj = cast(WDL.Expr.Array, arguments[0])
+            array_items = [str(self.get_expr(item)) for item in array_obj.items]
```
kinow, line 654: Fixed the mypy warning. Also, the arrow function caused a warning when validating the workflow, as that's not part of EC5

wdl2cwl/tests/wdl_files/tasks/select_all.wdl
```diff
@@ -0,0 +1,11 @@
...
+task test_select_all {
+  input {
+    Int one
+    Int? two
+  }
+  command {}
+  output {
+    Array[Int] first1 = select_all([one, two, 1, select_first([two, one])])
+  }
+}
```
kinow, line 11: Not used yet, but just to illustrate for now. Will update the unit test later to use this file (and fix the missing newline at EOF :grimacing: )

mr-c:
@kinow I ran your CWL code, but it always returns empty arrays for me. I had to make the following adjustment
of adding `return` before the element comparisons.

``` cwl
        outputEval: $([inputs.one, inputs.two, 1, [inputs.two, inputs.one].find(function(element)
            { return element !== null }) ].filter(function(element) { return element !== null })
            )
```

kinow:
For later, happy if anyone (@mr-c ? :grimacing: ) takes over :+1:

mr-c (approved):

PR #165, opened 2022-02-07
Refactor get_cwl_task_outputs()

### LINKED ISSUE #157: split `get_cwl_task_outputs` into separate functions for the separate uses
I think we are overloading `get_cwl_task_outputs` with too many tasks to accomplish by different callers of the function.

1. [`load_wdl_workflow()` line 233](https://github.com/common-workflow-lab/wdl-cwl-translator/blob/4aa8229bac7d1f42e48cae06229334398d5bb525/wdl2cwl/main.py#L233) building the CWL version of the workflow level outputs using the given the `effective_outputs()` object (`WDL.Env.Bindings[Decl]`) to get the following
- name of the workflow output (matching the WDL name of the workflow output, to be nice to the WDL users)
- CWL type for the WDL type of the output
- name of the step to get the output from
- name of the output from that step

2. [`load_wdl_workflow()` line 289](https://github.com/common-workflow-lab/wdl-cwl-translator/blob/4aa8229bac7d1f42e48cae06229334398d5bb525/wdl2cwl/main.py#L289) building the `cwl.WorkflowStepOutput` object using the current step (either a `WDL.Tree.Task` or `WDL.Tree.Workflow`)'s outputs (`List[WDL.Tree.Decl]`) to get the names (`str`) of the outputs

3. [`load_wdl_task()`](https://github.com/common-workflow-lab/wdl-cwl-translator/blob/4aa8229bac7d1f42e48cae06229334398d5bb525/wdl2cwl/main.py#L315) the nominal user of `get_cwl_task_outputs`; uses the entire list of `cwl.CommandOutputParameter` objects as-is to build the `cwl.CommandLineTool` from the `WDL.Tree.Task` outputs (`List[WDL.Tree.Decl]`)

In summary: three users of the same function, with three different sets of goals, and three different inputs :-)

I recommend we create three separate functions for these different purposes; there could be a helper function used by at least two of the new functions.

What do you think @Th3nn3ss @kinow ?

mr-c:
Fixes #157

### DIFF
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -4,7 +4,18 @@ import os
 import re
 import sys
 import textwrap
-from typing import Any, Dict, List, Optional, Sequence, Set, Tuple, Union, cast
+from typing import (
+    Any,
+    Dict,
+    List,
+    Optional,
+    Sequence,
+    Set,
+    Tuple,
+    Union,
+    cast,
+    Iterator,
+)
 
 import cwl_utils.parser.cwl_v1_2 as cwl
 import regex  # type: ignore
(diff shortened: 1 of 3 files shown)


wdl2cwl/main.py
```diff
@@ -152,6 +163,29 @@ def get_expr_name_with_is_file_check(wdl_expr: WDL.Expr.Ident) -> str:
     return expr_name if not is_file else f"{expr_name}.path"
 
 
+def get_workflow_outputs(
+    outputs: WDL.Env.Bindings[WDL.Type.Base],
+) -> Iterator[Tuple[str, Union[cwl.OutputArraySchema, str], str]]:
+    """Return the name, CWL type, and source for a workflow's effective_outputs()."""
+    for item in outputs:
+        output_name = item.name
+        meta_name = item.info.expr.expr.name[::-1].replace(".", "/", 1)[::-1]
```
Th3nn3ss, line 172: This is really cool and it totally works.

Th3nn3ss (approved):
I guess we need to have another function to handle workflow step inputs.

wdl2cwl/main.py
Th3nn3ss: The only issue I see with this is in the case of imported tasks. This meta_name doesn't reflect the name of the task it is being imported from. If this is not included the cwltool raises unidentified outputsource error.

Th3nn3ss (commented):
if len(item.expr.expr.referee.callee_id) == 2: 
                        meta_name = (
                           item.expr.expr.referee.callee_id[0] + "." + meta_name
                        )

wdl2cwl/main.py
mr-c: > The only issue I see with this is in the case of imported tasks. This meta_name doesn't reflect the name of the task it is being imported from. If this is not included the cwltool raises unidentified outputsource error.

This is the same code we had before, just in a different location. Can you provide a link to a WDL file that causes an issue?

Th3nn3ss: It's needed to solve the validation problem as in the case of `align_and_count.wdl`

Th3nn3ss: But that's why I approved it. I can make the change to pass validate. This is clearly a better way to handle workflow inputs an I did before with the get_cwl_outputs in PR #156

mr-c: Okay. Just confirming that you want me to merge as-in :-)

PR #167, opened 2022-02-07
Funcs improved

Th3nn3ss:
Little improvements to the `get_expr_apply` method and added some comments to the `get_outdir_requirment` method as some of the decisions made might seem a little confusing when read later out of context.


wdl2cwl/main.py
```diff
@@ -533,7 +542,7 @@ def get_expr_string(self, wdl_expr_string: WDL.Expr.String) -> str:
         for index, part in enumerate(parts[1:-1], start=1):
             if isinstance(
                 part,
-                (WDL.Expr.Placeholder, WDL.Expr.Apply, WDL.Expr.Get, WDL.Expr.Ident),
```
Th3nn3ss, line 536: Only placeholder expressions are used within WDL.Expr.Strings

wdl2cwl/main.py
```diff
@@ -561,6 +570,16 @@ def get_expr_ifthenelse(self, wdl_ifthenelse: WDL.Expr.IfThenElse) -> str:
 
     def get_expr_apply(self, wdl_apply_expr: WDL.Expr.Apply) -> str:
         """Translate WDL Apply Expressions."""
+        binary_ops = {
```
Th3nn3ss, line 573: Some operations are binary with just a left operand, right operand and an operator. This dictionary stores function names as keys and their operator symbol as values.

wdl2cwl/main.py
```diff
@@ -433,28 +433,37 @@ def get_outdir_requirement(
...
+                # runtime["disks"] usually have strings chars like local, HDD, SDD that are not needed by CWL
                 if (
                     isinstance(outdir, WDL.Expr.Apply)
                     and not outdir.function_name == "_add"
                 ):
+                    # If it contains an apply expr we don't want to process the _add function
+                    # that concatenates it to the chars in the string
                     expr_str = self.get_expr(outdir)
+                    # if the expr_str returns an integer that is declared as a static variable
+                    # return the result from the multiplication with 1024
```
mr-c, line 445: We can see that the result is multiplied by 1024, so don't repeat yourself; better to say why we multiply by 1024

wdl2cwl/main.py
```diff
@@ -433,28 +433,37 @@ def get_outdir_requirement(
                 isinstance(outdir, (WDL.Expr.Apply, WDL.Expr.String))
                 and outdir.literal is None
             ):
+                # runtime["disks"] usually have strings chars like local, HDD, SDD that are not needed by CWL
```
mr-c, line 436: I don't understand how this comment relates to any of the code that surrounds it

wdl2cwl/main.py
```diff
@@ -433,28 +433,37 @@ def get_outdir_requirement(
...
                     if isinstance(
                         obj, (WDL.Expr.Get, WDL.Expr.Apply, WDL.Expr.Placeholder)
                     ):
+                        # avoid python strings only WDL expressions are handled.
                         return self.get_outdir_requirement(obj)  # type: ignore
             elif isinstance(outdir, (WDL.Expr.Get, WDL.Expr.Placeholder)):
                 expr = self.get_expr(outdir)
                 expr_str = expr
                 if isinstance(outdir, (WDL.Expr.Placeholder)):
+                    # expr returns a string that contains $(). This has to be removed.
```
mr-c, line 466: `get_expr()` that returns a string wrapped in `$(` and `)`, yes?

Maybe `get_expr()` shouldn't and we just add the `$(` and `)` where it makes sense? Would be less mysterious :-)

wdl2cwl/main.py
mr-c: Great simplification! :+1:

mr-c (changes requested):

wdl2cwl/main.py
Th3nn3ss: Awesome Idea

PR #168, opened 2022-02-09
better error messages for malformed WDL

mr-c:


mr-c:
*sigh* ; @kinow  can you help me with https://github.com/common-workflow-lab/wdl-cwl-translator/runs/5128753953?check_suite_focus=true#step:7:153 ? (it reproduces locally for me, but I can't figure out what change introduced this problem; running the command manually has no error)

kinow:
Really interesting! We appear to have problems when `test_cli.py` runs before `test_cwl.py`. Something to do with the new `test_cli.py`. It looks like when the test makes our converter and miniwdl raise an error, then somehow the comments from your `parse_error.wdl` get persisted afterwards?

Trying to isolate and identify whether it's a problem in our code, or a bug in miniwdl. miniwdl uses asyncio under the hood to load the file and parse it, and we have pytest-xdist as well, so a bit tricky to have a clear way to reproduce and debug it. But fun! :nerd_face:

mr-c:
> Really interesting! We appear to have problems when `test_cli.py` runs before `test_cwl.py`. Something to do with the new `test_cli.py`. It looks like when the test makes our converter and miniwdl raise an error, then somehow the comments from your `parse_error.wdl` get persisted afterwards?
> 
That's a good hint, thanks! I think I can find a solution from here.

kinow:
> > Really interesting! We appear to have problems when `test_cli.py` runs before `test_cwl.py`. Something to do with the new `test_cli.py`. It looks like when the test makes our converter and miniwdl raise an error, then somehow the comments from your `parse_error.wdl` get persisted afterwards?
> 
> That's a good hint, thanks! I think I can find a solution from here.

Ah! But not is the fun part :smile: it appears to be a bug in the miniwdl comments cache: https://github.com/chanzuckerberg/miniwdl/blob/d50d97a9e263373d8d46610b2fdcc99b16059e49/WDL/_parser.py#L30

It is cleared only when the parser succeeds. When it fails the cache keeps the old comments (in this case, the `comment 1` and `comment 2` from that `parse_error.wdl`).

mr-c:
Yep, or we can manually call `WDL._parsers._lark_comments_buffer.clear()` in our `main.py` before calling the public miniwdl functions until the fixed miniwdl is release and we pin to that fixed version or newer

wdl2cwl/main.py
```diff
@@ -37,6 +38,7 @@ class ConversionException(Exception):
 def convert(doc: str) -> Dict[str, Any]:
     """Convert a WDL workflow, reading the file, into a CWL workflow Python object."""
     wdl_path = os.path.relpath(doc)
+    WDL._parser._lark_comments_buffer.clear()
```
kinow, line 41: :clap: good idea!

PR #173, opened 2022-02-13
Scatter with align and count

### LINKED ISSUE #146: basic WDL scatter functionality
https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#scatter

example:

https://github.com/broadinstitute/warp/blob/5593923ed2a1bcd9187887dd28e33fbeb72f41a5/pipelines/broad/annotation_filtration/AnnotationFiltration.wdl#L29

Th3nn3ss:
Fixes #146

Handle align_and_count_multiple_report.wdl that contains WDL.Tree.Scatter sections. 

`get_workflow_call` handles all calls in the body of `WDL.Tree.Workflow` and in the body of a `WDL.Tree.Scatter`
`get_workflow_scatter` handles all scatter objects in WDL.Tree.Workflow
`get_step_input_expr` was extended to handle `WDL.Expr.Get` expressions as well as handle `WDL.Expr.Get` with a `WDL.Tree.Scatter` referee object


mr-c:
Very exciting! Don't forget to pull in the functional test I wrote in https://github.com/common-workflow-lab/wdl-cwl-translator/commit/2f92ba11a2b062ac5ae3133f701547f33c320508#diff-4bd0bad20f775ef6cf25b77468f0cd495589a552373c70da0d7d5eaf62fb1e3dR22 (or I can do it, just tell me when)

wdl2cwl/main.py
```diff
@@ -309,99 +311,113 @@ def load_wdl_workflow(self, obj: WDL.Tree.Workflow) -> cwl.Workflow:
...
             outputs=outputs,
         )
 
+    def get_workflow_scatter(self, scatter: WDL.Tree.Scatter) -> List[cwl.WorkflowStep]:
+        """Get the CWL Workflow Step equivalent of a list of WDL Scatter Object."""
+        scatter_steps: List[cwl.WorkflowStep] = []
+        for body_obj in scatter.body:
+            if isinstance(body_obj, WDL.Tree.Call):
+                wp_step = self.get_workflow_call(body_obj)
+            wp_step.scatter = wp_step.in_[0].id
```
mr-c, line 349: I'm pretty sure this won't work for any step that is not `scatter`ing the first input.

mr-c:
Can you try with https://github.com/broadinstitute/warp/blob/5593923ed2a1bcd9187887dd28e33fbeb72f41a5/pipelines/broad/arrays/multi_sample/MultiSampleArrays.wdl  and then run `cwltool --validate` on the result?

wdl2cwl/main.py
mr-c: Here is proof

Make the following change to the order of the inputs of the `align_and_count` Task:

``` diff
diff --git a/wdl2cwl/tests/wdl_files/tasks/tasks_reports.wdl b/wdl2cwl/tests/wdl_files/tasks/tasks_reports.wdl
index 6e02e8c..b85b7d2 100644
--- a/wdl2cwl/tests/wdl_files/tasks/tasks_reports.wdl
+++ b/wdl2cwl/tests/wdl_files/tasks/tasks_reports.wdl
@@ -274,8 +274,8 @@ task fastqc {
 
 task align_and_count {
   input {
-    File   reads_bam
     File   ref_db
+    File   reads_bam
     Int    topNHits = 3
 
     Int?   machine_mem_gb
```

`make wdl2cwl/tests/cwl_files/align_and_count_multiple_report.cwl` and check the diff, this changes the `scatter` line:

``` diff
diff --git a/wdl2cwl/tests/cwl_files/align_and_count_multiple_report.cwl b/wdl2cwl/tests/cwl_files/align_and_count_multiple_report.cwl
index 7790762..257fe56 100644
--- a/wdl2cwl/tests/cwl_files/align_and_count_multiple_report.cwl
+++ b/wdl2cwl/tests/cwl_files/align_and_count_multiple_report.cwl
@@ -143,7 +143,7 @@ steps:
         baseCommand:
           - bash
           - script.bash
-    scatter: reads_bam
+    scatter: ref_db
   - id: align_and_count_summary
     in:
       - id: counts_txt
```

wdl2cwl/main.py
```diff
@@ -277,99 +282,117 @@ def load_wdl_workflow(self, obj: WDL.Tree.Workflow) -> cwl.Workflow:
...
+        """Get the CWL Workflow Step equivalent of a WDL Call Object."""
+        callee = call.callee
+        if callee is None:
+            raise ConversionException("callee is None. This should not be possible.")
+        cwl_callee_inputs = self.get_cwl_task_inputs(callee.inputs)
+        call_inputs = call.inputs
+        inputs_from_call: Dict[str, Tuple[str, Dict[str, Any]]] = {}
+        input_defaults = set()
+        if call_inputs:
+            if_scatter_handled = False
```
mr-c, line 334:
```suggestion
            scatter_handled = False
```
It is a little confusing to begin a variable with `if_`

wdl2cwl/main.py
```diff
@@ -277,99 +282,117 @@ def load_wdl_workflow(self, obj: WDL.Tree.Workflow) -> cwl.Workflow:
...
+                            },
+                        )
+                    else:
+                        input_expr = self.get_step_input_expr(value)  # type: ignore[arg-type]
+                        inputs_from_call[key] = (
+                            input_expr.replace(".", "/"),
+                            {},
+                        )
+                    if self.scatter_names and not if_scatter_handled:
+                        self.scatter_names[0] = key
```
mr-c, line 365: Why the first index?

mr-c (changes requested):

wdl2cwl/main.py
```diff
@@ -229,6 +229,7 @@ def __init__(self) -> None:
         """Initialize the sets used by the object and prevent inconsistent behaviours."""
         self.non_static_values: Set[str] = set()
         self.optional_cwl_null: Set[str] = set()
+        self.scatter_names: List[str] = []
```
mr-c, line 232: I don't understand this list. Is this a list of tasks that got scattered, the source of the scatter, the destination, or the newly created variable made as a result of WDL's `scatter` construct?

Th3nn3ss: It serves as a stack for scatter objects as they are encountered.

Th3nn3ss: When a scatter input is encountered it's destination is stored in the stack. The destination is stored is stored because we don't have access to the source withing the input object.

wdl2cwl/main.py
```diff
@@ -277,99 +282,117 @@ def load_wdl_workflow(self, obj: WDL.Tree.Workflow) -> cwl.Workflow:
...
+                            },
+                        )
+                    else:
+                        input_expr = self.get_step_input_expr(value)  # type: ignore[arg-type]
+                        inputs_from_call[key] = (
+                            input_expr.replace(".", "/"),
+                            {},
+                        )
+                    if self.scatter_names and not scatter_handled:
+                        self.scatter_names[-1] = key
```
Th3nn3ss, line 365: When a scatter input is encountered and the scatter's source has been appended to the scatter_names stack, it us updated with the destination which is the key. 

scatter_handled is updated to True so as to prevent the scatter_names from been updated when the call has more than one inputs. 

This implementation only handles one scatter input per scatter call.

wdl2cwl/main.py
```diff
@@ -277,99 +282,117 @@ def load_wdl_workflow(self, obj: WDL.Tree.Workflow) -> cwl.Workflow:
...
             outputs=outputs,
         )
 
+    def get_workflow_scatter(self, scatter: WDL.Tree.Scatter) -> List[cwl.WorkflowStep]:
+        """Get the CWL Workflow Step equivalent of a list of WDL Scatter Object."""
+        scatter_steps: List[cwl.WorkflowStep] = []
+        for body_obj in scatter.body:
+            if isinstance(body_obj, WDL.Tree.Call):
+                wp_step = self.get_workflow_call(body_obj)
+            wp_step.scatter = self.scatter_names.pop()
```
Th3nn3ss, line 320: Immediately a scatter call is returned as a wp_step the last element is popped off the stack and the scatter `wp_step.scatter` attribute is updated.

wdl2cwl/main.py
mr-c: What if there is a sub-workflow that has a scatter, would it conflict?

What about a workflow that has two separate scatters, what happens then?

Maybe another data structure that relates the scatter call name, the source, and the destination together would be safer?

wdl2cwl/main.py
```diff
@@ -277,99 +282,118 @@ def load_wdl_workflow(self, obj: WDL.Tree.Workflow) -> cwl.Workflow:
         wf_description = (
             obj.meta.pop("description") if "description" in obj.meta else None
         )
-        for call in obj.body:
-            if not isinstance(call, WDL.Tree.Call):
+        is_scatter_present = False
```
kinow, line 285: We could have a list for `requirements` here instead. Then where we set `scatter = True` we could just add the `ScatterRequirement` instead.

wdl2cwl/main.py
```diff
@@ -277,99 +282,118 @@ def load_wdl_workflow(self, obj: WDL.Tree.Workflow) -> cwl.Workflow:
...
         )
 
+    def get_workflow_scatter(self, scatter: WDL.Tree.Scatter) -> List[cwl.WorkflowStep]:
+        """Get the CWL Workflow Step equivalent of a list of WDL Scatter Object."""
+        scatter_steps: List[cwl.WorkflowStep] = []
+        for body_obj in scatter.body:
+            with WDLSourceLine(body_obj, ConversionException):
+                if isinstance(body_obj, WDL.Tree.Call):
+                    wf_step = self.get_workflow_call(body_obj)
+                wf_step.scatter = self.scatter_names.pop()
```
kinow, line 321: I think if the `isinstance` above fails, this would result in a... `NameNotFound` error? Can't recall the exception for variables not yet declared. But if the `if` above can be `False`, we have to be more defensive here IMO.

wdl2cwl/main.py
```diff
@@ -277,99 +282,118 @@ def load_wdl_workflow(self, obj: WDL.Tree.Workflow) -> cwl.Workflow:
...
+                scatter_steps.append(wf_step)
+        return scatter_steps
+
+    def get_workflow_call(self, call: WDL.Tree.Call) -> cwl.WorkflowStep:
+        """Get the CWL Workflow Step equivalent of a WDL Call Object."""
+        callee = call.callee
+        if callee is None:
+            raise ConversionException("callee is None. This should not be possible.")
```
kinow, line 329: Are we able to write a workflow that tests this?

wdl2cwl/main.py
kinow: Ah, ok, I think I understood how `scatter_names` is used. I think here we could empty the stack? Unless there's further need for `scatter_names`, or maybe make a copy at the beginning of the function, then empty as we process it. That would remove the need for the `scatter_handled` variable and make it easier to comprehend, I think.

wdl2cwl/main.py
Th3nn3ss: No. This is mostly because of mypy. It complains that callee could be None. which has never been encountered yet.

wdl2cwl/main.py
Th3nn3ss: I am yet to encounter anything other than a WDL.Tree.Call inside a scatter object but I believe it's possible. The `with WDLSourceLine` handles the case when it is not.

wdl2cwl/main.py
Th3nn3ss: If the parent of the call is a scatter object then the `scatter_names` stack will still be used there. The last name entered into the `scatter_names` stack is popped at get_workflow_scatter.

wdl2cwl/main.py
mr-c: we would get multiple `ScatterRequirement`s that way, if different steps in the same workflow had a scatter

mr-c (approved):

PR #179, opened 2022-02-14
read_{tsv,int} ; added more stdout support

### LINKED ISSUE #178: read_*(stdout())
Nice example at https://github.com/DataBiosphere/topmed-workflows/blob/209e5b1b56abbf900717dc4c2d08e2093e7efdda/aligner/functional-equivalence-wdl/FunctionalEquivalence.wdl#L300

mr-c:
print warning but don't break if we are unable to extract a static
docker name

Fixes #178

### DIFF
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -164,7 +164,7 @@ def get_input(input_name: str) -> str:
 
 def get_cwl_docker_requirements(
     wdl_docker: Union[WDL.Expr.Get, WDL.Expr.String]
-) -> cwl.ProcessRequirement:
+) -> Optional[cwl.ProcessRequirement]:
     """Translate WDL Runtime Docker requirements to CWL Docker Requirement."""
     if isinstance(wdl_docker, WDL.Expr.String) and wdl_docker.literal:
         dockerpull = get_literal_value(wdl_docker)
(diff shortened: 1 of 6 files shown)


wdl2cwl/main.py
```diff
@@ -373,26 +400,27 @@ def load_wdl_workflow(self, obj: WDL.Tree.Workflow) -> cwl.Workflow:
...
             doc=description,
             inputs=cwl_inputs,
+            outputs=None,
             hints=hints,
             requirements=requirements,
-            outputs=cwl_outputs,
             cwlVersion="v1.2",
             baseCommand=["bash", "script.bash"] if run_script else ["true"],
         )
+        tool.outputs = self.set_cwl_task_outputs(obj.outputs, obj.parameter_meta, tool)
```
mr-c, line 420: so we can set `stdout` to a name of our choosing, if need be

kinow: Oh, I thought it was weird to create the `tool` object, then pass it to that function. Then reading the code I saw the `glob_str = tool.stdout = "_stdout"`. Tried to think other ways to have it so we didn't need the function being called after the `tool` is created... but couldn't come up with anything not too messy :laughing: 

Thanks for adding the comment here :+1:

kinow (approved):

PR #181, opened 2022-02-15
implement read_lines() with a functional test

### LINKED ISSUE #175: implement read_lines
https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#arraystring-read_linesstringfile

Example https://github.com/biowdl/tasks/blob/c5cfd2f5acc2ff729987b86d38b29af046677fdc/pacbio.wdl#L83

mr-c:
also fix read_tsv() and add functional test

Fixes #175

### DIFF
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -237,15 +237,19 @@ read_funcs = {
   if (contents == 'false') { return false;}
   throw "'read_boolean' received neither 'true' nor 'false': " + self[0].contents;
 }""",
+    "read_lines": r"""${
+  var contents = self[0].contents.replace(/\r\n$/, "").replace(/\n$/, "").replace(/\r$/, "");
+  // ^ remove any trailing newline to prevent a null being returned
+  return contents.split(/\r\n|\r|\n/);
+}""",
     "read_tsv": r"""${
-  var result;
-  self.contents.split(/\n|(\n\r)/).forEach(function(line) {
-    var line_array;
-    line.split('\t').forEach(function(field) {
-      line_array.push(field)
-    })
-    result.push(line_array);
-  })
+  var result = Array();
+  var contents = self[0].contents.replace(/\r\n$/, "").replace(/\n$/, "").replace(/\r$/, "");
+  // ^ remove any trailing newline to prevent a null being returned
+  contents.split(/\r\n|\r|\n/).forEach(function(line) {
+    result.push(line.split('\t'));
+  });
+  return result;
 }""",
 }
 
@@ -763,6 +767,7 @@ class Converter:
             "read_int",
             "read_boolean",
             "read_tsv",
+            "read_lines",
         }
         function_name = wdl_apply_expr.function_name
         arguments = wdl_apply_expr.arguments
--- a/wdl2cwl/tests/cwl_tests.yaml
+++ b/wdl2cwl/tests/cwl_tests.yaml
@@ -166,3 +166,16 @@
   tool: cwl_files/literal_test.cwl
   output:
     result: "true 42 6.022 hoopla README.md true,false 42,23 6.022,10,23 Hello,World README.md,LICENSE"
+
+- doc: read_lines(stdout)
+  tool: cwl_files/pacbio.cwl#ccsChunks
+  job: inputs/pacbio_ccsChunks.yaml
+  output:
+    chunks: ["1/10", "2/10", "3/10", "4/10", "5/10", "6/10", "7/10", "8/10", "9/10", "10/10"]
+
+- doc: read_tsv() test
+  tool: cwl_files/Utilities.cwl#CreateSequenceGroupingTSV
+  job: inputs/Utilities_CreateSequenceGroupingTSV.yaml
+  output:
+    sequence_grouping: [ [ "22:1+" ], [ "X:1+" ], [ "Y:1+" ] ]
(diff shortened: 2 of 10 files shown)


mr-c:
@kinow I think we need a new reviewing rule: no new javascript without functional CWL test compared to the miniwdl output :-)

wdl2cwl/main.py
```diff
@@ -236,16 +236,20 @@ def get_literal_value(expr: WDL.Expr.Base) -> Optional[Any]:
...
   if (contents == 'false') { return false;}
   throw "'read_boolean' received neither 'true' nor 'false': " + self[0].contents;
+}""",
+    "read_lines": r"""${
+  var contents = self[0].contents.replace(/\r\n$/, "").replace(/\n$/, "").replace(/\r$/, "");
+  // ^ remove any trailing newline to prevent a null being returned
+  return contents.split(/\r\n|\r|\n/);
 }""",
     "read_tsv": r"""${
-  var result;
```
mr-c, line 241: wasn't initialized

wdl2cwl/main.py
```diff
@@ -236,16 +236,20 @@ def get_literal_value(expr: WDL.Expr.Base) -> Optional[Any]:
...
-  self.contents.split(/\n|(\n\r)/).forEach(function(line) {
```
mr-c, line 242: the regex used capturing groups  `(  )` instead of the OR operator `|`. For split, the capturing groups are returned separately

wdl2cwl/main.py
```diff
@@ -236,16 +236,20 @@ def get_literal_value(expr: WDL.Expr.Base) -> Optional[Any]:
...
-    var line_array;
-    line.split('\t').forEach(function(field) {
-      line_array.push(field)
-    })
```
mr-c, line 246: No need to use `forEach` here as we don't further modify or have any logic around the results; `split()` already returns an array

wdl2cwl/tests/cwl_tests.yaml
```diff
@@ -166,3 +166,16 @@
...
+  job: inputs/pacbio_ccsChunks.yaml
+  output:
+    chunks: ["1/10", "2/10", "3/10", "4/10", "5/10", "6/10", "7/10", "8/10", "9/10", "10/10"]
+
+- doc: read_tsv() test
+  tool: cwl_files/Utilities.cwl#CreateSequenceGroupingTSV
+  job: inputs/Utilities_CreateSequenceGroupingTSV.yaml
+  output:
+    sequence_grouping: [ [ "22:1+" ], [ "X:1+" ], [ "Y:1+" ] ]
+    sequence_grouping_with_unmapped: [ [ "22:1+" ], [ "X:1+" ], [ "Y:1+" ], [ "unmapped" ] ]
```
mr-c, line 181:
```
2022-02-15 10:32:46.267 wdl.t:CreateSequenceGroupingTSV done
{
  "outputs": {
    "CreateSequenceGroupingTSV.sequence_grouping_with_unmapped": [
      [
        "22:1+"
      ],
      [
        "X:1+"
      ],
      [
        "Y:1+"
      ],
      [
        "unmapped"
      ]
    ],
    "CreateSequenceGroupingTSV.sequence_grouping": [
      [
        "22:1+"
      ],
      [
        "X:1+"
      ],
      [
        "Y:1+"
      ]
    ]
  },
  "dir": "/home/michael/wdl2cwl/20220215_103243_CreateSequenceGroupingTSV"
}
```

mr-c (commented):
Proof of the matching miniWDL output

kinow:
(will have during lunch break in some hours :+1: )

wdl2cwl/main.py
```diff
@@ -236,16 +236,20 @@ def get_literal_value(expr: WDL.Expr.Base) -> Optional[Any]:
   if (contents == 'true') { return true;}
   if (contents == 'false') { return false;}
   throw "'read_boolean' received neither 'true' nor 'false': " + self[0].contents;
+}""",
+    "read_lines": r"""${
+  var contents = self[0].contents.replace(/\r\n$/, "").replace(/\n$/, "").replace(/\r$/, "");
```
kinow, line 241: Looks good to me. Do you know what's the `self` variable above? That's JS right? So not the Python `self`... I couldn't find it looking around this line, so more for my curiosity :) 

Thanks!

kinow (approved):

wdl2cwl/main.py
mr-c: https://www.commonwl.org/v1.2/CommandLineTool.html#Parameter_references

>  The following parameters must be provided:
>
>    `inputs`: The input object to the current Process.
>    `self`: A context-specific value. The contextual values for `self` are documented for specific fields elsewhere in this specification. If a contextual value of 'self' is not documented for a field, it must be 'null'.
>    `runtime`: An object containing configuration details. Specific to the process type. An implementation may provide opaque strings for any or all fields of runtime. These must be filled in by the platform after processing the Tool but before actual execution. Parameter references and expressions may only use the literal string value of the field and must not perform computation on the contents, except where noted otherwise.

These particular expressions end up in the `outputEval` field, so we look at https://www.commonwl.org/v1.2/CommandLineTool.html#CommandOutputBinding to learn that

> If `glob` was specified, the value of `self` must be an array containing file objects that were matched. If no files were matched, `self` must be a zero length array; if a single file was matched, the value of `self` is an array of a single element.

kinow: Thanks!!

PR #184, opened 2022-02-15
Fix quoting & Add another example of if-then-else

### LINKED ISSUE #74: WDL If-then-else expressions are not handled
Currently, If-then-else expressions in commands are not handled by the translator. (https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#true-and-false)

A workflow using if-then-else expressions is given in this workflow (https://github.com/biowdl/tasks/blob/bc1bacf11498d2d30b85591cfccdcf71ef0966a5/whatshap.wdl#L52)

### LINKED ISSUE #131: handle advanced quoting in expressions in the WDL command block
https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#strings

See https://github.com/biowdl/tasks/blob/bc1bacf11498d2d30b85591cfccdcf71ef0966a5/whatshap.wdl#L52

mr-c:
Closes #74 
Fixes #131

### DIFF
--- a/wdl2cwl/main.py
+++ b/wdl2cwl/main.py
@@ -723,28 +723,18 @@ class Converter:
         """Translate WDL String Expressions."""
         if wdl_expr_string.literal is not None:
             return str(wdl_expr_string.literal)
-        string = ""
         parts = wdl_expr_string.parts
-        for index, part in enumerate(parts[1:-1], start=1):
-            if isinstance(
-                part,
-                WDL.Expr.Placeholder,
-            ):
-                placeholder = self.get_expr(part)
-                part = (
-                    "" if parts[index - 1] == '"' or parts[index - 1] == "'" else "' + "  # type: ignore
-                )
-                part += placeholder
-                part += (
-                    "" if parts[index + 1] == '"' or parts[index + 1] == "'" else " + '"  # type: ignore
-                )
-            string += part
-        # condition to determine if the opening and closing quotes should be added to string
-        # for cases where a placeholder begins or ends a WDL.Expr.String
-        if type(parts[1]) == str:
-            string = "'" + string
-        if type(parts[-2]) == str:
-            string = string + "'"
+        q = cast(str, parts[0])[0]
+        string = (
+            f"{q}{parts[1]}{q}"
+            if isinstance(parts[1], str)
+            else self.get_expr(parts[1])
+        )
+        if parts[2:-1]:
+            string += " + " + " + ".join(
+                f"{q}{part}{q}" if isinstance(part, str) else self.get_expr(part)
+                for part in parts[2:-1]
+            )
         return string
 
     def get_expr_ifthenelse(self, wdl_ifthenelse: WDL.Expr.IfThenElse) -> str:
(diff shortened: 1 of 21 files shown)


wdl2cwl/tests/cwl_files/FunctionalEquivalence.cwl
```diff
@@ -99,7 +99,7 @@ $graph:
       - class: ResourceRequirement
         coresMin: 16
         ramMin: 13351.4404296875
-        outdirMin: $((inputs.disk_size.replace("\..*", "") ) * 1024)
+        outdirMin: $(parseFloat(inputs.disk_size.replace("\\..*", "") ) * 1024)
```
mr-c, line 102: `outdirMin` needs a `Float` or `Int`, but `replace()` returns a string

https://github.com/common-workflow-lab/wdl-cwl-translator/blob/6961aa80933fd97a773f05fa2ad0de864f3c0a01/wdl2cwl/tests/wdl_files/FunctionalEquivalence.wdl#L131

wdl2cwl/tests/cwl_files/vardict.cwl
```diff
@@ -10,7 +10,7 @@ requirements:
             set -e -o pipefail
             export JAVA_OPTS="-Xmx$(inputs.javaXmx) -XX:ParallelGCThreads=1"
             vardict-java \
-            -th  $(inputs.threads) \
+            $("-th " + inputs.threads) \
```
kinow, line 13: The other values below are in a different format, `-PARAM $(expr)`, rather than this one `("-PARAM" + expr)`. Is there are reason for one format over the other?

wdl2cwl/main.py
```diff
@@ -725,29 +722,19 @@ def get_expr(self, wdl_expr: WDL.Expr.Base) -> str:
     def get_expr_string(self, wdl_expr_string: WDL.Expr.String) -> str:
         """Translate WDL String Expressions."""
         if wdl_expr_string.literal is not None:
-            return f'"{wdl_expr_string.literal.value}"'
-        string = ""
+            return str(wdl_expr_string.literal)
```
kinow, line 725: @mr-c couldn't we call `return str(get_literal_value(wdl_expr_string))` here? That way we wouldn't have to call `get_literal_value` in the code directly, but instead centralize string translations in this `get_expr_string`.

kinow (approved):
Couple comments, but no blockers. Feel free to resolve discussion & merge if the comments do not apply :+1: 

Thanks!

wdl2cwl/tests/cwl_files/vardict.cwl
mr-c: Good question. This is the same pattern in the original WDL

https://github.com/common-workflow-lab/wdl-cwl-translator/blob/97a24489bb57804a659dabba38ef4bd64949293a/wdl2cwl/tests/wdl_files/vardict.wdl#L62

wdl2cwl/main.py
mr-c: Feel free to try it and let me know? My memory is that the result is different. This way preserves the original quotes, if I remember correctly

kinow: Moved to follow-up issue so it doesn't block this PR: https://github.com/common-workflow-lab/wdl-cwl-translator/issues/185

:+1:

PR #186, opened 2022-02-16
improve handling of placeholder options

### LINKED ISSUE #114: Simplify translate_wdl_placeholder

mr-c:
Fixes #114 

- `defined()` now works outside of WDL Placeholder strings
- `default` as a placeholder string option is now implemented

### DIFF
--- a/wdl2cwl/tests/cwl_files/FunctionalEquivalence.cwl
+++ b/wdl2cwl/tests/cwl_files/FunctionalEquivalence.cwl
@@ -549,7 +549,7 @@ $graph:
           - entryname: script.bash
             entry: |4
 
-                python -c "print $()"
+                python -c "print $(inputs.sizes.join("+"))"
       - class: InlineJavascriptRequirement
       - class: NetworkAccess
         networkAccess: true
--- a/wdl2cwl/tests/cwl_files/InternalTasks.cwl
+++ b/wdl2cwl/tests/cwl_files/InternalTasks.cwl
@@ -67,8 +67,8 @@ $graph:
                   --HAPLOTYPE_MAP $(inputs.haplotype_database_file.path) \
                   --EXPECTED_GENOTYPING_PLATFORMS FLUIDIGM \
                   --EXPECTED_GENOTYPING_PLATFORMS GENERAL_ARRAY \
-                  $(inputs.ignoreSpecificGenotypesLsid ? inputs.ignoreSpecificGenotypesLsid === null ? "" : "--IGNORE_SPECIFIC_GENOTYPES_LSID \"" + inputs.ignoreSpecificGenotypesLsid + "\"" : "") \
-                  $(inputs.ignoreSpecificGenotypesPlatform ? inputs.ignoreSpecificGenotypesPlatform === null ? "" : "--IGNORE_SPECIFIC_GENOTYPES_PLATFORM \"" + inputs.ignoreSpecificGenotypesPlatform + "\"" : "") \
+                  $(inputs.ignoreSpecificGenotypesLsid !== null ? inputs.ignoreSpecificGenotypesLsid === null ? "" : "--IGNORE_SPECIFIC_GENOTYPES_LSID \"" + inputs.ignoreSpecificGenotypesLsid + "\"" : "") \
+                  $(inputs.ignoreSpecificGenotypesPlatform !== null ? inputs.ignoreSpecificGenotypesPlatform === null ? "" : "--IGNORE_SPECIFIC_GENOTYPES_PLATFORM \"" + inputs.ignoreSpecificGenotypesPlatform + "\"" : "") \
                   --MERCURY_FP_STORE_URI $MERCURY_FP_STORE_URI \
                   --CREDENTIALS_VAULT_PATH $MERCURY_AUTH_KEY \
                   --ERR_NO_GENOTYPES_AVAILABLE 7
--- a/wdl2cwl/tests/cwl_files/deepvariant.cwl
+++ b/wdl2cwl/tests/cwl_files/deepvariant.cwl
@@ -25,7 +25,7 @@ requirements:
             $("--regions " + inputs.regions) \
             $(inputs.sampleName === null ? "" : "--sample_name " + inputs.sampleName) \
             $(inputs.postprocessVariantsExtraArgs === null ? "" : "--postprocess_variants_extra_args " + inputs.postprocessVariantsExtraArgs) \
-            $(inputs.VCFStatsReport === null ? "--novcf_stats_report" : "--vcf_stats_report")
+            $(inputs.VCFStatsReport === null ? "" : inputs.VCFStatsReport ? "--vcf_stats_report" : "--novcf_stats_report")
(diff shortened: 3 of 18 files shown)


wdl2cwl/tests/cwl_files/InternalTasks.cwl
```diff
@@ -67,8 +67,8 @@ $graph:
                   --HAPLOTYPE_MAP $(inputs.haplotype_database_file.path) \
                   --EXPECTED_GENOTYPING_PLATFORMS FLUIDIGM \
                   --EXPECTED_GENOTYPING_PLATFORMS GENERAL_ARRAY \
-                  $(inputs.ignoreSpecificGenotypesLsid ? inputs.ignoreSpecificGenotypesLsid === null ? "" : "--IGNORE_SPECIFIC_GENOTYPES_LSID \"" + inputs.ignoreSpecificGenotypesLsid + "\"" : "") \
-                  $(inputs.ignoreSpecificGenotypesPlatform ? inputs.ignoreSpecificGenotypesPlatform === null ? "" : "--IGNORE_SPECIFIC_GENOTYPES_PLATFORM \"" + inputs.ignoreSpecificGenotypesPlatform + "\"" : "") \
+                  $(inputs.ignoreSpecificGenotypesLsid !== null ? inputs.ignoreSpecificGenotypesLsid === null ? "" : "--IGNORE_SPECIFIC_GENOTYPES_LSID \"" + inputs.ignoreSpecificGenotypesLsid + "\"" : "") \
+                  $(inputs.ignoreSpecificGenotypesPlatform !== null ? inputs.ignoreSpecificGenotypesPlatform === null ? "" : "--IGNORE_SPECIFIC_GENOTYPES_PLATFORM \"" + inputs.ignoreSpecificGenotypesPlatform + "\"" : "") \
```
mr-c, line 71: a bit redundant, yes, just like the original WDL (they don't need `defined()`
https://github.com/common-workflow-lab/wdl-cwl-translator/blob/63d60b51421a68b819411e36ed47a1df547db411/wdl2cwl/tests/wdl_files/InternalTasks.wdl#L96

wdl2cwl/tests/cwl_files/FunctionalEquivalence.cwl
```diff
@@ -549,7 +549,7 @@ $graph:
           - entryname: script.bash
             entry: |4
 
-                python -c "print $()"
+                python -c "print $(inputs.sizes.join("+"))"
```
mr-c, line 552: One example of missing output from prior code

wdl2cwl/tests/cwl_files/deepvariant.cwl
```diff
@@ -25,7 +25,7 @@ requirements:
             $("--regions " + inputs.regions) \
             $(inputs.sampleName === null ? "" : "--sample_name " + inputs.sampleName) \
             $(inputs.postprocessVariantsExtraArgs === null ? "" : "--postprocess_variants_extra_args " + inputs.postprocessVariantsExtraArgs) \
-            $(inputs.VCFStatsReport === null ? "--novcf_stats_report" : "--vcf_stats_report")
+            $(inputs.VCFStatsReport === null ? "" : inputs.VCFStatsReport ? "--vcf_stats_report" : "--novcf_stats_report")
```
mr-c, line 28: Now we properly skip adding `--vcf_stats_report` or `--novcf_stats_report` if `VCFStatsReport` is unset

wdl2cwl/tests/cwl_files/placeholder_options_test.cwl
```diff
@@ -0,0 +1,50 @@
...
+  - id: result
+    type: string
+    outputBinding:
+        loadContents: true
+        glob: _stdout
+        outputEval: $(self[0].contents.replace(/[\r\n]+$/, ''))
+  - id: present
+    type: boolean
+    outputBinding:
+        outputEval: $(inputs.absent !== null)
```
mr-c, line 50: translation of a `defined()` not inside a WDL Placeholders string that would have been incorrectly translated previously as `inputs.absent`

wdl2cwl/tests/cwl_files/InternalTasks.cwl
Th3nn3ss: This looks very interesting. So each time the `defined( )` function is used it will checked to see if it's `null` how is it different from when it's used in placeholder options? Do you have a link to a wdl file that has this functionality?

Th3nn3ss (approved):
Nice work with the nice_quote function. It really captured a fear I had before. 
I always thought there was more to the defined function, this implementation does handle it properly.

mr-c:
> Nice work with the nice_quote function. It really captured a fear I had before. I always thought there was more to the defined function, this implementation does handle it properly.

Thanks! I actually [want to get rid of `nice_quote`](https://github.com/chanzuckerberg/miniwdl/issues/551) since I was [able to remove the need elsewhere](https://github.com/common-workflow-lab/wdl-cwl-translator/commit/48a1569d0645b164bfaa7ef13faf8474612b9d00#diff-b5a65c2befbbd0b6453805272729b622f41b2be683f8258432e08f4370332265L728)

PR #187, opened 2022-02-16
Enforce that Array[Type]+ are non-empty

mr-c:
https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#arrayx
> An `Array` may have an empty value (i.e. an array of length zero), unless it is declared using `+`, the non-empty postfix quantifier, which represents a constraint that the `Array` value must contain one-or-more elements. 

Only enforced for WDL Task inputs for now.

For Workflow inputs we will have to synthesize a new step to do the checking 

Uses a hack from https://github.com/common-workflow-language/common-workflow-language/issues/907


wdl2cwl/tests/cwl_files/align_and_count_multiple_report.cwl
```diff
@@ -205,6 +205,9 @@ steps:
         baseCommand:
           - bash
           - script.bash
+        arguments:
+          - valueFrom: ${if (inputs.counts_txt.length == 0) {throw "counts_txt must
+                contain at least one item.";} else { return "";}}
```
mr-c, line 210: Total hack. Doesn't change the command line if the array has at least one item. Throws a javascript exception if the array is empty, effectively avoiding execution of the tool if the input is malformed.

Th3nn3ss (approved):
Awesome

PR #201, opened 2022-06-02
Handle cases where the WDL workflow output contains exprs

### LINKED ISSUE #200: 'Apply' object has no attribute 'expr'
my wdl file :
https://github.com/admi2u/blast_wdl/blob/master/blast.wdl#L74

while convert this file to cwl use wdl2cwl script, report this error:
`Traceback (most recent call last):
  File "/root/wdl-cwl-translator/wdl2cwl/main.py", line 726, in get_workflow_outputs
        self,
    output_source = item_expr.expr.name[::-1].replace(".", "/", 1)[::-1]
AttributeError: 'Apply' object has no attribute 'expr'`

how can i fix this problem?

kinow:
Closes #200 

@admi2u, can you check if this branch's code works for you? It produced the `blast.cwl` file in this branch/pull request.

I am taking a look at the tests that broke after this change, and will push a new commit to fix it later.

Thanks
Bruno


wdl2cwl/tests/cwl_files/blast.cwl
```diff
@@ -0,0 +1,217 @@
...
+      - 'null'
+  - id: runtblastx.max_hsps
+    type:
+      - int
+      - 'null'
+steps: []
+outputs:
+  - id: blast.fina_output
+    outputSource: '[inputs["runtblastx.out"] === null '
+    type: File
```
kinow, line 217: Resulting CWL file :point_up:

wdl2cwl/tests/wdl_files/blast.wdl
```diff
@@ -0,0 +1,325 @@
...
+        Queryfa    = queryfa,
+        docker     = blast_docker,
+        outfmt     = outfmt,
+        evalue     = evalue,
+        Outfile    = Outfile,
+        threads    = threads
+    }
+  }
+  output {
+    File fina_output =select_first([runtblastx.out,runblastp.out,runblastn.out,runblastx.out,runtblastn.out])
```
mr-c, line 76: And it wasn't even a complete (mis)translation 😅

mr-c (changes requested):

wdl2cwl/tests/wdl_files/blast.wdl
kinow: It looks like I cannot call `get_expr` for all expressions, as it messed up the `outputSoruce` values in other files, breaking some tests.

I changed it to handle `Expr.Get` as before, but for others to call `get_expr`. Still not looking 100%, since the expression appears to be partially evaluated? :confused: Pushed the latest CWL file and changes.

wdl2cwl/tests/cwl_files/blast.cwl
kinow: Right, IIUC, the `select_first` is returning a string for an JS expression, but the `outputSource` is only `string` or `string[]`. I think we could replace that output by an `outputBinding` and `outputEval`, but I think we don't have `outputEval` at workflow level.

The WDL Conditionals were also not translated. I think we would have to handle the conditionals and then move this `select_first` to a step of the workflow? Then use `outputSource: step_created.fina_output`?

mr-c: > The WDL Conditionals were also not translated. I think we would have to handle the conditionals and then move this select_first to a step of the workflow? Then use outputSource: step_created.fina_output?

Yes! Probably can do both in the same step

admi2u:
@kinow Thank you very much. I am sorry for late reply.
I have tested this branch, the results are consistent with your test results, the `outputSource` expression is wrapped in `= f"$({ ... })".

wdl2cwl/tests/cwl_files/blast.cwl
```diff
@@ -0,0 +1,335 @@
...
+      default: false
+      type: boolean
+  baseCommand:
+    - bash
+    - script.bash
+  outputs:
+    - id: out
+      type: File
+      outputBinding:
+        glob: $(inputs.Outfile)
```
kinow, line 335: @mr-c , I tried to manually partially translate the `blast.wdl` file. Could you take a look at the file above and see if it's going in the right direction here, please?

There are a couple TODO's where I am not sure how to translate something from WDL, :pray: 

I used two tasks only, translating them both into `CommandLineTool`'s. Also used that `$graph` trick to pack multiple processes in the same file. Finally, added `Workflow` named `main` for the `blast.wdl` workflow definition. The `Workflow` uses conditional to pick the first non null (i.e. `select_first` and `if`'s from WDL).

wdl2cwl/tests/cwl_files/blast.cwl
```diff
@@ -0,0 +1,335 @@
...
+  id: main
+  inputs:
+    - id: blast_docker_override
+      type:
+        - string
+        - 'null'
+    - id: blast_docker
+      type: string
+      # TODO: how to translate the original?
+      #       String blast_docker = select_first([blast_docker_override,"swr.cn-south-1.myhuaweicloud.com/cngbdb/blast:1.2"])
```
kinow, line 13: :point_up:

wdl2cwl/tests/cwl_files/blast.cwl
```diff
@@ -0,0 +1,335 @@
...
+    - id: outfmt
+      type: int
+      default: 7
+    - id: evalue
+      type: float
+      default: 10
+    # TODO: how to translate the original?
+    #       String Outfile = basename(queryfa)+'.blast_result.txt'
+    - id: Outfile
+      type: string
```
kinow, line 31: :point_up:

mr-c:
@kinow, when you've got a working and tested manually written CWL conversion, please see (or take over) my work at https://github.com/common-workflow-lab/wdl-cwl-translator/pull/191/files for synthesizing new steps to deal with workflow output expressions

wdl2cwl/tests/cwl_files/blast.cwl
mr-c: For the general case of programmatic conversion, we would need to synthesize a new step to compute the usable value of `blast_docker` and use the output of that step everywhere that originally referred to the `blast_docker` input.

However, if I was converting this by hand I would have combined `blast_docker_override` and `blast_docker` into a single CWL input with a default value. However, this is being used to override the docker image name with a dynamic value, which is [not yet supported in CWL](https://github.com/common-workflow-lab/wdl-cwl-translator#incompatibilities-possibly-requiring-manual-intervention); so it would be useless :-P

Anyhow, we still need a general way to convert WDL expressions in the definition of inputs. I recommend creating a separate issue for that; modifying this WDL to not need workflow input declaration expressions, and focusing on synthesizing new steps for workflow output expressions.

wdl2cwl/tests/cwl_files/blast.cwl
```diff
@@ -0,0 +1,335 @@
...
+  steps:
+    - id: runblastp
+      in:
+        Fname: fname
+        Queryfa: queryfa
+        docker: blast_docker
+        outfmt: outfmt
+        Outfile: Outfile
+        threads: threads
+      run: runblastp
```
mr-c, line 133:
```suggestion
      run: #runblastp
```

mr-c (changes requested):

wdl2cwl/tests/cwl_files/blast.cwl
mr-c: This is another WDL expression in the definition of an input; I've made an issue for that https://github.com/common-workflow-lab/wdl-cwl-translator/issues/240

